PR #89, opened 2026-02-26
Fix reentrant lock crashes in IMAP handler paths

jdudley:
## Summary\n- avoid re-entrant lock usage in fetch handlers\n- harden command/logging lock paths against callback re-entry\n\n## Validation\n- swift build\n- swift test -q\n\n## Context\nAddresses NIOLock precondition crashes observed under real IMAP callback interleavings.


odrobnik:
@jdudley Thanks for the PR. I don't quite understand why we need this, but I merged it nevertheless hoping that it doesn't hurt...

PR #97, opened 2026-02-28
Fix crash in decodeQuotedPrintable on truncated input

tabmail-kmyi:
## Summary

- Fixes a crash in `decodeQuotedPrintable()` when `=` appears near the end of input (e.g. `"Hello=X"`, `"Hello="`)
- Wires up `decodeQuotedPrintableLossy()` as a fallback in `MessagePart.textContent` for graceful degradation

## Root Cause

`index(_:offsetBy:limitedBy:)` returns `endIndex` (not `nil`) when the offset lands exactly on the limit. Using that index in a closed range subscript (`nextIndex...endIndex`) reads past the buffer, causing a `String index out of range` crash.

The `decodeQuotedPrintableLossy()` method added in #76f10df correctly handles these cases, but the strict `decodeQuotedPrintable()` still crashes on truncated input.

## Fix

Replace `index(_:offsetBy:limitedBy:)` with two sequential `index(after:)` + `< endIndex` guards that are correct by construction.

Also add lossy fallback in `MessagePart.textContent` (both the base64-decoded and direct UTF-8 paths) so malformed QP bodies return degraded content instead of `nil`.

## Test plan

- [x] `swift build` succeeds
- [x] All 70 tests pass (`swift test`)
- [x] New test cases cover: `"Hello=X"`, `"Hello="`, `"Hello=A"` — all previously crashed

### DIFF
--- a/Sources/SwiftMail/Extensions/String+QuotedPrintable.swift
+++ b/Sources/SwiftMail/Extensions/String+QuotedPrintable.swift
@@ -56,8 +56,11 @@ extension String {
 
             if char == "=" {
                 let nextIndex = withoutSoftBreaks.index(after: index)
-                guard nextIndex < withoutSoftBreaks.endIndex,
-                      let nextNextIndex = withoutSoftBreaks.index(nextIndex, offsetBy: 1, limitedBy: withoutSoftBreaks.endIndex) else {
+                guard nextIndex < withoutSoftBreaks.endIndex else {
+                    return nil
+                }
+                let nextNextIndex = withoutSoftBreaks.index(after: nextIndex)
+                guard nextNextIndex < withoutSoftBreaks.endIndex else {
                     return nil
                 }
                 let hex = String(withoutSoftBreaks[nextIndex...nextNextIndex])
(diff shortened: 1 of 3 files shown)


odrobnik:
I verified the unit-test additions in PR #97 against the current main branch.

The proposed test changes appear to already be present on main — the patch does not apply cleanly because equivalent assertions are already in QuotedPrintableTests.

I also ran the tests on current main, and they pass both for the targeted QuotedPrintableTests and for the full test suite.

Given recent SwiftMail changes to quoted-printable handling (transfer decoding now happens directly on Data bytes first), this PR may now be outdated or superseded, with less reliance on the earlier string-based path.

tabmail-kmyi:
Thanks for verifying @odrobnik! You're right — the fix and tests are already on main. The QP decoding now uses safe `index(after:)` + boundary guards, and the edge-case tests (`"Hello=X"`, `"Hello="`, `"Hello=A"`) are covered in `edgeCasesAndMalformedInput()`. With transfer decoding moving to the Data path, this PR is fully superseded. Closing.

PR #114, opened 2026-03-07
Expose IMAP threading headers in fetch metadata

jdudley:
## Summary
- capture BODY[HEADER] and RFC822.HEADER literal streams in FetchMessageInfoHandler
- parse In-Reply-To and References into MessageInfo.additionalFields
- preserve the existing dynamic XOAUTH2 token reauth support required by KestrelMail
- add single and bulk fetch tests covering threading header extraction

## Testing
- swift test --filter FetchMessageInfoHandlerTests --package-path /tmp/SwiftMail-header-threading
- swift test --filter SwiftIMAPTests --package-path /tmp/SwiftMail-header-threading

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift
@@ -11,6 +11,9 @@ import NIOConcurrencyHelpers
 final class FetchMessageInfoHandler: BaseIMAPCommandHandler<[MessageInfo]>, IMAPCommandHandler, @unchecked Sendable {
     /// Collected email headers
     private var messageInfos: [MessageInfo] = []
+    private var currentSequenceNumber: SequenceNumber?
+    private var currentHeaderLiteral = Data()
+    private var collectingThreadingHeaders = false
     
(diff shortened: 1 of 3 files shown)


odrobnik (commented):
Thanks for this @jdudley — good work on the streaming header parsing and tests! A few observations:

### Threading headers

1. **`In-Reply-To` is already extracted from ENVELOPE** — the existing handler (line ~174) parses `envelope.inReplyTo` into `messageInfo.inReplyTo`. The raw header extraction here duplicates that into `additionalFields["in-reply-to"]`, which creates two different access paths for the same data. I'd suggest skipping `In-Reply-To` in `extractThreadingHeaders()` since it's already a first-class property.

2. **`References` deserves a dedicated property** — rather than burying it in `additionalFields["references"]`, it would be cleaner as `messageInfo.references: String?` on `MessageInfo`, matching the pattern of `inReplyTo`. That way threading info has a consistent API surface.

### Scope

3. **Second commit is unrelated** — the XOAUTH2 dynamic token provider (`setXOAUTH2AccessTokenProvider`) is a separate feature. Could you split it into its own PR? Keeps review and git history cleaner.

### Conflict

4. There's a **merge conflict** with `main`. Could you rebase on current `main` to resolve it?

Tests look solid 👍

PR #120, opened 2026-03-08
Add SMTP SIZE support

jdudley:
## Summary
- expose parsed SMTP SIZE capability data on SMTPServer
- compute exact message octet counts from the generated MIME content on Email
- include SIZE=<octets> in MAIL FROM when sending mail
- add focused SMTP tests for SIZE parsing, command formatting, and message-size calculation

## Motivation
Some SMTP servers advertise a maximum accepted message size via the RFC 1870 SIZE extension. Exposing that capability and announcing the actual MIME payload size during MAIL FROM lets clients make better send decisions and align with server expectations without hard-coding provider-specific limits.

## Validation
- swift test --filter SMTPTests

### DIFF
--- a/Sources/SwiftMail/Extensions/Email+Content.swift
+++ b/Sources/SwiftMail/Extensions/Email+Content.swift
@@ -1,6 +1,10 @@
 import Foundation
 
 extension Email {
+    public func messageSizeOctets(use8BitMIME: Bool = false) -> Int {
+        constructContent(use8BitMIME: use8BitMIME).utf8.count
+    }
+
     /**
      Build the MIME encoded email body.
 
--- a/Sources/SwiftMail/Extensions/Email+Content.swift
+++ b/Sources/SwiftMail/Extensions/Email+Content.swift
@@ -1,8 +1,18 @@
 import Foundation
 
 extension Email {
+    struct PreparedContent {
+        let content: String
+        let messageSizeOctets: Int
+    }
+
+    func preparedContent(use8BitMIME: Bool = false) -> PreparedContent {
+        let content = constructContent(use8BitMIME: use8BitMIME)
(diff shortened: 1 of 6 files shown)


odrobnik (changes requested):
Good PR — clean, well-scoped, and the `MailFromCommand` refactor is nice. A few issues to address:

### 1. SIZE sent unconditionally (RFC 1870 violation)

Currently `messageSizeOctets` is always passed to `MailFromCommand`, even when the server didn't advertise the `SIZE` capability. Per [RFC 1870 §4](https://datatracker.ietf.org/doc/html/rfc1870#section-4), clients SHOULD NOT send `SIZE` unless the server announced it. The `messageSizeOctets` parameter should only be set when `maximumMessageSizeOctets != nil` (i.e., server advertised `SIZE`).

### 2. No early rejection when message exceeds server limit

`maximumMessageSizeOctets` is parsed from capabilities but never checked. If the server says `SIZE 10485760` and the email is 20MB, we should fail early with a clear error (e.g. `SMTPError.messageTooLarge`) rather than sending the full payload and getting a `552` rejection.

### 3. Wasteful double content construction

`messageSizeOctets(use8BitMIME:)` calls `constructContent().utf8.count`, building the entire MIME body in memory just to count bytes. Then `sendEmail` builds it again. For large emails with attachments this doubles memory usage. Consider computing the size and content together, or caching the result.

### 4. Confusing variable name

`announcedMessageSizeOctets` reads like the server's announced limit, but it's actually the email's size. Something like `emailSizeOctets` would be clearer.

### 5. Minor: public capability properties

`supports8BitMIME` and `maximumMessageSizeOctets` are public but only meaningful after EHLO. Fine for now, but worth a doc comment noting that.

odrobnik (approved):
All review feedback addressed cleanly. LGTM ✅

PR #121, opened 2026-03-08
feat: allow pre-set Message-Id via additionalHeaders

tabmail-kmyi:
## Summary

- When `additionalHeaders` contains a `Message-Id` (or `Message-ID`) key, `constructContent()` uses that value instead of auto-generating a UUID
- The preset key is excluded from the additionalHeaders iteration loop to prevent duplicate `Message-Id` headers
- If no preset is provided, behavior is unchanged (auto-generates `<UUID@domain>`)

## Motivation

Email clients that use SMTP for sending and IMAP APPEND for saving to the Sent folder need the **same** `Message-ID` in both operations. Without this, each call to `constructContent()` generates a different UUID, making it impossible to deduplicate on retry (e.g. after a connection drop between SMTP send and IMAP APPEND).

This is standard behavior in email clients — the Message-ID is generated once and reused across both the delivery (SMTP) and storage (IMAP APPEND) paths.

## Usage

```swift
var email = Email(sender: sender, recipients: recipients, subject: "Hello", textBody: "World")
email.additionalHeaders = ["Message-Id": "<my-stable-id@example.com>"]

// Both calls produce the same Message-Id header
let smtpContent = email.constructContent(use8BitMIME: true)
let imapContent = email.constructContent(use8BitMIME: true)
```

## Test plan

- [x] `testConstructContentAutoGeneratesMessageId` — no preset → auto-generates as before
- [x] `testConstructContentUsesPresetMessageId` — preset `Message-Id` is used, only one occurrence
- [x] `testConstructContentUsesPresetMessageIDUpperCase` — `Message-ID` (uppercase) variant works
- [x] `testConstructContentStableMessageIdAcrossCalls` — same preset produces identical headers across calls
- [x] All existing tests pass

### DIFF
--- a/Sources/SwiftMail/Core/Models/Email.swift
+++ b/Sources/SwiftMail/Core/Models/Email.swift
@@ -31,8 +31,20 @@ public struct Email: Sendable {
     /** Optional attachments for the email */
     public var attachments: [Attachment]?
     
-    /** Optional additional headers (e.g. Message-Id, X-Custom-Header) */
+    /** Optional additional headers (e.g. X-Custom-Header) */
     public var additionalHeaders: [String: String]?
+
+    /**
+     Optional Message-ID for this email.
+
+     When set, `constructContent()` uses this value instead of auto-generating one.
+     This allows callers to use the same Message-ID across both SMTP send and
+     IMAP APPEND (e.g. saving to the Sent folder), which is required for
+     deduplication on retry after connection drops or app crashes.
+
+     When `nil`, a Message-ID is auto-generated during content construction.
+     */
+    public var messageID: MessageID?
     
     /**
      Initialize a new email with EmailAddress objects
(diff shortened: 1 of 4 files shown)


tabmail-kmyi:
Updated — `messageID` is now a `MessageID` custom type:

- **`MessageID` struct** — `localPart` + `domain`, `CustomStringConvertible` for angle-bracket formatting, `Sendable` + `Hashable`
- **`.generate(domain:)`** factory for auto-generation
- **Failable `init(_:)`** parser that handles both `<id@domain>` and `id@domain` formats, rejects empty parts
- **`Email.messageID`** is now `MessageID?`
- **`constructContent()`** uses `self.messageID ?? .generate(domain:)`
- Tests updated + 4 new tests for `MessageID` itself (generate, parse valid, parse without brackets, parse invalid)

odrobnik:
Thank you for your contributions!

PR #128, opened 2026-03-11
Add UID EXPUNGE support

jdudley:
## Summary
- add `UID EXPUNGE` support to `IMAPServer` and `IMAPNamedConnection`
- route the command through `IMAPConnection` using the `UID` command form
- use targeted `UID EXPUNGE` in the UID-based move fallback when `MOVE` is unavailable but `UIDPLUS` is present
- add coverage for wire format and capability gating

## Why
`UID EXPUNGE` is part of `UIDPLUS` and allows callers to permanently remove only the selected `\Deleted` messages by UID instead of expunging the entire selected mailbox. Exposing it directly is useful on its own, and using it in the UID-based move fallback avoids broad `EXPUNGE` behavior on servers that support `UIDPLUS` but not `MOVE`.

## Validation
- `swift test`

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Commands/ServerCommands.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Commands/ServerCommands.swift
@@ -109,4 +109,26 @@ struct ExpungeCommand: IMAPTaggedCommand {
     func toTaggedCommand(tag: String) -> TaggedCommand {
         return TaggedCommand(tag: tag, command: .expunge)
     }
-} 
+}
+
+/// Command for expunging specific deleted messages by UID.
+struct UIDExpungeCommand: IMAPTaggedCommand {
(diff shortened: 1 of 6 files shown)


odrobnik (changes requested):
The `move` fallback safety improvement here is excellent — `COPY + STORE + UID EXPUNGE` is vastly safer than a global `EXPUNGE`.

One API design request: Instead of introducing `uidExpunge(messages:)` as a distinct method name, let's overload `expunge` directly:

```swift
/// Permanently removes specific deleted messages by UID when UIDPLUS is available.
public func expunge(messages identifierSet: UIDSet) async throws
```

Since the parameter is strongly typed to `UIDSet` (which is `MessageIdentifierSet<UID>`), it's clear at compile time that this is the UID version. This keeps the API surface symmetric with the existing `expunge()`.

PR #129, opened 2026-03-11
Honor provided Message-ID headers without duplication

jdudley:
## Summary
- honor `additionalHeaders["Message-ID"]` as a compatibility fallback when `email.messageID` is unset
- avoid re-emitting the provided header a second time during additional header serialization
- preserve raw `Message-ID` pass-through when the additional-header value is not parseable as a typed `MessageID`
- add tests for exact-once behavior, typed-property precedence, case-insensitive matching, and the unparseable-header case

## Why
PR #121 allowed callers to pre-set a `Message-Id` via `additionalHeaders`, but current `main` still writes the generated or typed `Message-Id` header first and then appends all additional headers verbatim. If a caller uses `additionalHeaders["Message-ID"]` without `email.messageID`, the serialized message can still contain duplicate `Message-Id` headers.

This keeps the newer `email.messageID` API as the preferred path while preserving the documented older behavior for existing callers, including raw additional-header pass-through when the value does not round-trip through the typed `MessageID` parser.

## Validation
- `swift test --filter SMTPTests`

### DIFF
--- a/Sources/SwiftMail/Extensions/Email+Content.swift
+++ b/Sources/SwiftMail/Extensions/Email+Content.swift
@@ -43,13 +43,22 @@ extension Email {
 
         content += "Subject: \(self.subject)\r\n"
         content += "Date: \(Self.rfc2822Date())\r\n"
-        let msgID = self.messageID ?? resolvedMessageIDHeader() ?? .generate(domain: Self.senderDomain(from: self.sender))
-        content += "Message-Id: \(msgID)\r\n"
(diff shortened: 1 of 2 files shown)


odrobnik (approved):
This is a great compatibility bridge. The precedence rules (property > parseable header > unparseable header > auto-gen) are exactly right, and the tests cover the edge cases nicely. Thanks for bringing this back! LGTM ✅

PR #141, opened 2026-03-24
feat: add IMAP command pipelining for body part fetches

tabmail-kmyi:
## Summary

Adds support for sending multiple `FETCH BODY[section]` commands without awaiting individual responses, per RFC 3501 §5.5. This significantly improves body fetch throughput by eliminating per-command round-trip latency.

- **`PipelinedCommandDispatcher`**: NIO channel handler that routes responses to multiple in-flight command handlers by tag. Maintains insertion order for correct untagged FETCH response routing (IMAP processes commands in order).
- **`PipelinedFetchPartHandler`**: Lightweight handler collecting streamed body-part data per pipelined command. Managed by the dispatcher, not added to the NIO pipeline directly.
- **`executePipelinedFetchParts`** on `IMAPConnection`: Sends N fetch-part commands under a single `commandQueue` lock, collects results via the dispatcher. No interleaving with other commands.
- **`fetchPartsPipelined`** on `IMAPServer`/`IMAPNamedConnection`: Public API.

The entire pipelined batch holds the `commandQueue` lock, preventing interleaving with other commands (SELECT, LOGIN, etc.). This is safe because pipelined commands are all FETCH operations on the same already-selected mailbox.

### Performance

Real-world testing with an IMAP server (15-connection limit, 14 active connections):

| Metric | Before | After |
|--------|--------|-------|
| Sub-batch of 100 UIDs | 4-6 seconds | 2-3 seconds |
| Throughput | ~25 bodies/sec | ~37 bodies/sec |
| Improvement | — | **~50% faster** |

Server-side command processing time is the remaining bottleneck — the pipeline eliminates client-side serial round-trip overhead.

## Test plan

- [x] `PipelinedFetchPartHandler`: data collection, idempotent fail, streaming bytes, ignore after finish
- [x] `PipelinedCommandDispatcher`: pending count, per-command failure isolation, completion promise
- [x] Full test suite passes (222 tests)
- [x] Tested on live IMAP server with 14 parallel connections

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
@@ -1,0 +1,111 @@
+// PipelinedCommandDispatcher.swift
+// NIO channel handler that routes responses to multiple in-flight pipelined command handlers.
+//
+// IMAP RFC 3501 §5.5 allows clients to send multiple commands without waiting for responses.
+// The server processes commands in order and sends tagged responses (A001 OK, A002 OK, etc.)
+// so responses can be matched to commands by tag. Untagged responses (e.g., * FETCH data)
+// always belong to the oldest pending command (server processes commands serially).
+//
+// This handler sits in the NIO pipeline during a pipelined batch. It maintains an ordered
+// registry of (tag → PipelinedHandler) and routes responses accordingly.
+
+import Foundation
+import Logging
+@preconcurrency import NIOIMAP
+import NIOIMAPCore
+import NIO
+import NIOConcurrencyHelpers
+
+final class PipelinedCommandDispatcher: ChannelInboundHandler, RemovableChannelHandler, @unchecked Sendable {
+    typealias InboundIn = Response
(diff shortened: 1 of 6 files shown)


odrobnik (commented):
Good implementation of RFC 3501 §5.5 command pipelining. The dispatcher design is clean — routing by tag for tagged responses, oldest-pending for untagged FETCHes.

**Strengths:**
- Thread safety with `NIOLock` throughout
- Double-resolve guards prevent NIO "leaking promise" fatal errors
- Proper error propagation (BYE, channel inactive, timeout)
- Clean separation: dispatcher routes, handlers collect

**Minor suggestion:**
In `executePipelinedFetchParts`, this:
```swift
for (_, (tag, uid, section)) in tagToRequest.enumerated() {
```
Could be simplified to:
```swift
for (tag, uid, section) in tagToRequest {
```
Since the outer index isn't used (it's captured separately in the `futures[i]` loop).

**Question:** Is there a reason to keep partial results if some fetches fail? Currently it collects all successes and logs failures. That seems reasonable for bulk operations — caller gets what succeeded. Just want to confirm that's intentional vs. fail-fast on first error.

PR #156, opened 2026-04-29
Fix SMTP LOGIN auth: dispatch lifecycle methods to subclasses

devidb:
## Summary

`LoginAuthHandler` defines `channelRegistered`, `handlerAdded`, `channelActive`, `handlerRemoved` to capture the `ChannelHandlerContext` into a stored property, but `BaseSMTPHandler` does not implement these methods. The protocol witness table for any subclass of `BaseSMTPHandler` therefore points at the `ChannelInboundHandler` default implementations, and methods declared on the subclass without `override` silently never run.

Result: when SMTP servers require `AUTH LOGIN`, the credential closure fires with `self.context == nil` and the promise fails with `connectionFailed("Channel context is nil")`.

## Reproduction

Connect to any SMTP server that advertises `AUTH LOGIN` (e.g. typical IMAP/SMTP providers: Mailo, Infomaniak, OVH, many self-hosted setups). Call `loginAuth(username:password:)`. The promise fails immediately when SwiftMail tries to send the base64-encoded username — `self.context` is nil.

## Fix

- **`BaseSMTPHandler.swift`** — implement the four lifecycle methods with sensible defaults (no-op for `handlerAdded`/`handlerRemoved`, `fireXxx` propagation for `channelRegistered`/`channelActive`). This places them in the witness table, so subclasses can override them.
- **`LoginAuthHandler.swift`** — mark the four methods `override public` so the subclass versions actually get dispatched.

The same latent bug would affect any other subclass of `BaseSMTPHandler` that overrides these lifecycle methods; this fix prevents it for all subclasses going forward.

## Test plan

- [ ] Verify SMTP `AUTH LOGIN` against a server that requires it (e.g. Mailo)
- [ ] Verify SMTP `AUTH PLAIN` still works (no regression)
- [ ] Verify normal IMAP flows (which do not depend on this code path)

### DIFF
--- a/Sources/SwiftMail/SMTP/Handlers/BaseSMTPHandler.swift
+++ b/Sources/SwiftMail/SMTP/Handlers/BaseSMTPHandler.swift
@@ -103,14 +103,35 @@ class BaseSMTPHandler<T: Sendable>: ChannelInboundHandler, RemovableChannelHandl
     public func errorCaught(context: ChannelHandlerContext, error: Error) {
         // Fail the promise with the error
         promise.fail(error)
-        
+
         // Remove this handler from the pipeline
         context.pipeline.removeHandler(self, promise: nil)
-        
+
         // Forward the error to the next handler
         context.fireErrorCaught(error)
     }
-    
+
+    // These four lifecycle methods are implemented here so subclasses can `override` them.
+    // Without an implementation on this class, the protocol witness table for any subclass
+    // points at the `ChannelInboundHandler` default implementations, and methods declared
+    // on the subclass (without `override`) silently never get called.
+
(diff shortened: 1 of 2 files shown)


odrobnik (approved):
Reviewed the diff and tested locally. The lifecycle methods added to BaseSMTPHandler make LoginAuthHandler's overrides participate in NIO protocol dispatch, so the handler context is populated before AUTH LOGIN challenges need to write credentials. The default BaseSMTPHandler lifecycle behavior also preserves event propagation for the other SMTP command handlers.

Verification: `swift test` passed in a detached PR worktree with 255 tests passing. The only warnings observed were existing IMAP Sendable warnings unrelated to this PR.

PR #160, opened 2026-05-09
Add slim/UID-flags fetches, mailbox delete & rename, RFC822.SIZE, named US zones

smbrownai:
## Summary

Adds a handful of related capabilities the IMAP layer was missing — all tested in production against a heterogeneous set of accounts (Fastmail, Namecheap Private Email, iCloud).

### New fetch commands
- **`FetchSlimMessageInfoCommand`** — header-only fetch that drops `BODYSTRUCTURE` and the full header section. Requests `UID + ENVELOPE + INTERNALDATE + FLAGS + RFC822.SIZE` plus a small set of named header fields (`List-Unsubscribe`, `List-Unsubscribe-Post`, `List-ID`, `Auto-Submitted`, `Precedence`) — common newsletter / auto-mail signals. Per-message responses are roughly an order of magnitude smaller than `FetchMessageInfoCommand`. Important for large mailboxes where the full FETCH response would exceed the 10-second per-command timeout.
- **`FetchUIDFlagsCommand`** — `(UID FLAGS)` only. Smallest possible per-message payload (~50 bytes). Designed for incremental sync diffing where you only want to know which UIDs are still on the server and their flag state.
- Both come with bulk + streaming convenience methods on `IMAPServer` and per-command chunk-size constants (`slimFetchChunkSize = 500`, `uidFlagsFetchChunkSize = 5000`) sized for their payload weight.

### Mailbox delete and rename
- **`IMAPServer.deleteMailbox(_:)`** and **`IMAPServer.renameMailbox(from:to:)`** fill out the mailbox-management surface alongside the existing `createMailbox(_:)`.
- Errors map to new `IMAPError.deleteFailed` / `renameFailed` cases with full `description` / `failureReason` coverage.

### `MessageInfo.size`
- New `size: Int?` field, populated from `RFC822.SIZE` via a new `case .rfc822Size` arm in `FetchMessageInfoHandler.updateHeader(_:with:)`.

### `BODY[HEADER.FIELDS (…)]` parsing
- `FetchMessageInfoHandler.shouldCollectThreadingHeaders` now also collects `.headerFields(...)` responses, not just `.header`. The existing parsing path stores everything not already in ENVELOPE into `additionalFields`, so `HEADER.FIELDS`-targeted fetches just work without any further changes.

### Named US time zones in date parsing
- Builds on top of the recent permissive-date-parsing work in #158. `parseEnvelopeDate` now also handles RFC 5322 §4.3 obsolete named US time zones (`PST`, `PDT`, `EST`, `EDT`, `CST`, `CDT`, `MST`, `MDT`, plus `AKST`/`AKDT`, `HST`/`HDT`, `UT`/`GMT`/`UTC`). Previously these failed silently because `DateFormatter`'s `Z` token doesn't recognise named zones — most large historical mailboxes have plenty of these.
- Implementation is a single preprocessing pass that rewrites a trailing named zone token to its numeric offset before format trial. Out-of-range fields still fail strictly.

## Test plan

- [x] Cullapse (the app this came out of) runs the full test suite of slim, UID+flags, delete, and rename against three live IMAP servers — all pass.
- [x] Slim fetch on a 23K-message mailbox completes in ~3 minutes via the streaming variant; previously the bulk variant timed out at the 10s limit.
- [x] UID+FLAGS fetch on a 23K-message mailbox returns in <2 seconds.
- [x] Named-zone date strings (`Wed, 31 Jan 2018 15:02:22 EST`, etc.) now parse correctly; previously fell into the warning branch.
- [ ] No new tests added in this PR — happy to add some focused on the date-parser additions if you'd like.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Commands/DeleteMailboxCommand.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Commands/DeleteMailboxCommand.swift
@@ -1,0 +1,26 @@
+import Foundation
+import NIO
+import NIOIMAPCore
+
+/** Command to delete an existing mailbox. */
+struct DeleteMailboxCommand: IMAPTaggedCommand {
+    typealias ResultType = Void
+    typealias HandlerType = DeleteMailboxHandler
+
+    let mailboxName: String
+
+    init(mailboxName: String) {
+        self.mailboxName = mailboxName
+    }
+
+    func validate() throws {
+        guard !mailboxName.isEmpty else {
+            throw IMAPError.invalidArgument("Mailbox name must not be empty")
+        }
+    }
+
+    func toTaggedCommand(tag: String) -> TaggedCommand {
+        let mailbox = MailboxName(ByteBuffer(string: mailboxName))
+        return TaggedCommand(tag: tag, command: .delete(mailbox))
+    }
+}
--- a/Sources/SwiftMail/IMAP/IMAP/Commands/FetchSlimMessageInfoCommand.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Commands/FetchSlimMessageInfoCommand.swift
@@ -1,0 +1,67 @@
+// FetchSlimMessageInfoCommand.swift
+// Like FetchMessageInfoCommand but skips BODYSTRUCTURE and the full header section. Requests
+// only UID, ENVELOPE, INTERNALDATE, FLAGS, RFC822.SIZE, plus a small set of named header fields
+// useful for newsletter / auto-mail detection. Per-message responses are roughly an order of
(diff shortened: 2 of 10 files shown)


odrobnik:
@smbrownai there are some review comments by CoPilot which look like they should be addressed.

Question: couldn't we unify the internal fetch commands which have the same return type under a single command that is passed an enum like FetchProfile, e.g. `.full`, .`simple` -> then then determines which values to request from the fetch?

odrobnik:
The parts without the slim fetching have been extracted and merged via #164 - discussion about the slim fetch implementation should happen here #163 - I would like to extend the existing API rather then add tons of new one.

PR #161, opened 2026-05-16
Adopt SwiftLint with auto-fix pre-commit hook and CI

odrobnik:
## Summary

- Adopt SwiftLint as the lint gate: in-repo pre-commit hook + GitHub Actions workflow + `.swiftlint.yml` configured for this codebase.
- Clean up 1927 existing violations down to **0** across 273 files. `swiftlint lint --strict` is clean and `swift test` passes 281/281.
- Large core types refactored via extension splits (the SwiftLint field-notes pattern): `IMAPServer.swift` 2431 → 217 lines (18 extension files), `IMAPConnection.swift` 1367 → 230 lines (7 extension files), `SMTPServer.swift` 1017 → 370 lines (4 extension files), plus splits in `String+MIME`, `String+QuotedPrintable`, `EMLParser`, `IMAPNamedConnection`, `IMAPTestServer`, the IMAP CLI demo, and several long test files.

## What's in the infrastructure

- **`scripts/hooks/pre-commit`** — lints **staged Swift files only** (otherwise existing violations in untouched files would block every commit). Auto-fix first, then re-stage, then `--strict` blocks on anything left. Refuses to run when a staged file also has unstaged edits — the single most important safety rail against `swiftlint --fix` + `git add` silently sweeping working-tree changes into the commit.
- **`scripts/install-hooks.sh`** — one-time per-clone: `git config core.hooksPath scripts/hooks`.
- **`.github/workflows/swiftlint.yml`** — macOS runner + `brew install swiftlint`, for parity with developer's local hook (same SwiftLint version, no drift). Path-filtered so doc-only PRs don't run lint.
- **`.swiftlint.yml`** — includes `Sources`/`Tests`/`Demos`; excludes `.build`/`.docc`. The only rule customization is a surgical `identifier_name.excluded` list for `to`, `cc`, `os`, `on`, `or` — RFC 5322/2971/3501 protocol names that appear on the public API surface where renaming would break consumers or lose protocol-level meaning.

## What's in the cleanup

Three buckets per the field notes:

1. **Auto-fixable (~1466 violations)** — trailing whitespace, syntactic sugar, trailing commas, implicit optional init, etc., via `swiftlint --fix`.
2. **Mechanical (~382 violations)** — identifier renames (`fd` → `fileDescriptor`, `kb`/`mb`/`gb` → `kilobytes`/`megabytes`/`gigabytes`, `s`/`q`/`ct`/`hi`/`lo` → meaningful names), line wrapping with care for `Logger.Message` arguments (which need a `String` `let` binding plus interpolation since `+` concatenation breaks the protocol), switch-case dedents (one pass per nesting level), `.data(using:.utf8)!` → `Data(_.utf8)`, removed `try!`/`as!` where the result was already guarded by an explicit type/error check.
3. **Structural (~79 violations)** — split large types via extensions. The 249-line `idle(on:configuration:)` orchestration in IMAPServer was tamed by extracting a non-isolated `IMAPResilientIdleRunner` enum so the cycle logic can run in a `Task.detached` while still being splittable into per-phase static helpers. Large enum switches collapsed into `static let map: [Key: Value]` lookups where possible (e.g. MIME-type-by-extension in `Attachment`, name-based mailbox special-use detection in `Mailbox`).

## Surgical exemptions

A few `// swiftlint:disable` comments remain with one-line reasons:

- **`force_try`** on compile-time-constant regex literals in `IMAPLogger` — pattern is a literal, cannot fail at runtime.
- **`force_cast`** in `CommandHandler.swift` where the cast is guarded by a preceding `ResultType.self == Void.self` check, and in `MessageIdentifierSet.swift` where each cast is preceded by an explicit type-pair check.
- **`optional_data_string_conversion`** on lossy `String(decoding:as:)` sites in `IMAPServer`, `Mailbox`, `String+QuotedPrintable`, `SendContentCommand` — these handle raw email bytes where preserving the data (with replacement characters) matters more than failing on non-UTF-8 input.
- **`function_parameter_count`** on `LogHandler.log(...)` overrides in the CLI demos — the 7-parameter signature is dictated by the `LogHandler` protocol.
- **`nesting`** on `Mailbox.Info.Attributes` — public API uses the nested path; flattening would break consumers.

## Test plan

- [x] `swiftlint lint --strict` reports `0 violations, 0 serious in 273 files`
- [x] `swift build` clean (Sources, Demos, Tests)
- [x] `swift test` passes `281/281` tests
- [x] `git config core.hooksPath` set after `scripts/install-hooks.sh`
- [ ] Reviewer: try a sample commit with a `--fix`-able Swift change to confirm the hook fires and auto-fixes
- [ ] Reviewer: try a sample commit with a real lint violation to confirm `--strict` blocks
- [ ] Reviewer: confirm CI workflow runs on this PR (`SwiftLint` check should be green)

## Notes for the reviewer

- This is intentionally a large refactor. The extension-based splits are mechanical (no behavior changes) but cross dozens of files. Diff-by-file rather than diff-by-line is the easier way to read it.
- A handful of `private` storage on actors/classes was widened to `internal` so extension files could reach it. No public API surface changed.
- The pre-commit hook lints staged files only — so even if a future commit somehow introduces a structural violation in an untouched file, the hook won't block unrelated commits. CI catches that case.


odrobnik:
Superseded by #166, which keeps the same SwiftLint adoption goal but rebuilds the history cleanly on top of current main — one commit per logical step so the lint count drops are reviewable, CI green throughout, and the macOS hang this PR was bisecting is identified and worked around.

PR #176, opened 2026-06-10
Send text/calendar invites inline per RFC 6047 so clients render Accept/Decline UI

hjbrandt:
## Problem

SwiftMail can already *parse* `text/calendar` parts on the receive side, but on the send side a calendar invite attached to an `Email` goes out like any other regular attachment: base64-encoded with `Content-Disposition: attachment`. Mail clients (Apple Mail, Outlook, Gmail, iCloud) only render their Accept/Decline invite UI when the iMIP (RFC 6047 §2.2) part ships as readable text with an inline (or absent) disposition — so invites sent through SwiftMail degrade to a plain `.ics` file attachment.

## What this does

Two changes to `Email+Content.swift`, mirroring how the major providers format outgoing invites:

1. **Single-invite messages become `multipart/alternative`.** When the message carries exactly one UTF-8 `text/calendar` attachment and nothing else (no HTML body, no inline attachments), the body is built as `multipart/alternative`: the `text/plain` body plus the ICS as 7bit text. This is the shape clients key the invite UI off — `multipart/mixed` with identical parts does not trigger it.

2. **In `multipart/mixed`, calendar attachments ship inline as 7bit text** (`Content-Disposition: inline`), while all other attachments keep the existing base64 behavior. Non-UTF-8 calendar data falls back to the regular base64 path.

The ICS is normalized to CRLF line endings before being emitted verbatim — producers commonly emit bare LF, which would otherwise violate SMTP DATA framing (RFC 5321 §2.3.8) and defeat dot-stuffing in `SendContentCommand`.

The 7bit label follows RFC 6047's examples; re-encoding (base64/quoted-printable) is precisely what breaks invite recognition in several clients.

## Tests

Four new cases in `SMTPTests`: the alternative shape for a single invite, the inline calendar part in a mixed message, bare-LF → CRLF normalization, and the non-UTF-8 base64 fallback. Full suite passes locally (292 tests).

## Field history

This formatting has been running for months as a local patch on a SwiftMail-based CLI sending invites through Proton Bridge — verified rendering Accept/Decline UI in Apple Mail, Outlook, Gmail, and iCloud.

### DIFF
--- a/Sources/SwiftMail/Extensions/Email+Content.swift
+++ b/Sources/SwiftMail/Extensions/Email+Content.swift
@@ -107,6 +107,23 @@ extension Email {
         let hasInline = !self.inlineAttachments.isEmpty
         let hasRegular = !self.regularAttachments.isEmpty
 
+        // iMIP invite shortcut (RFC 6047 §2.2): when the message carries exactly one
+        // text/calendar part and nothing else, ship it as multipart/alternative —
+        // text/plain body alternative + the ICS. Mail clients (Apple Mail, Outlook,
+        // Gmail, iCloud) key their Accept/Decline UI off this shape;
+        // multipart/mixed with the same parts does not trigger it.
+        if !hasHtmlBody && !hasInline,
+           self.regularAttachments.count == 1,
+           let icsText = calendarInviteText(for: self.regularAttachments[0]) {
+            writeCalendarAlternative(
+                context: context,
+                invite: self.regularAttachments[0],
+                icsText: icsText,
+                into: &content
+            )
+            return
+        }
+
         if hasRegular {
             writeMultipartMixed(context: context, hasHtmlBody: hasHtmlBody, hasInline: hasInline, into: &content)
         } else if hasHtmlBody && hasInline {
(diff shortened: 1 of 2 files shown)


odrobnik (commented):
## Review

The approach is right: the `multipart/alternative` shape for a lone invite and the inline 7bit calendar part in `multipart/mixed` match how the major providers format outgoing iMIP, and the field history is reassuring. I checked the MIME framing details and ran the suite locally — all 57 SMTP tests pass, including the four new ones. Two findings below should be addressed before merge; both are in the same helper and have a small, in-idiom fix.

### What I verified

- **CRLF normalization** (`calendarInviteText`): the three-step replace handles CRLF, bare CR, and bare LF in the correct order. ✔️
- **Boundary framing** (`terminatedICSBody`): both the trailing-newline and no-trailing-newline cases yield `…END:VCALENDAR\r\n` + blank line + boundary, consistent with the file's existing part convention. ✔️
- **Dot-stuffing claim**: confirmed `SendContentCommand.dotStuff` handles leading dots at the transport layer, so verbatim ICS text is safe on that front. ✔️
- **Part ordering**: text/plain first, text/calendar last — correct for `multipart/alternative` (last renderable part wins), so capable clients prefer the invite. ✔️
- **Non-UTF-8 fallback** to base64 is the right escape hatch and is tested. ✔️

### Finding 1 — non-ASCII ICS ships as 8-bit data labeled `7bit`, ignoring `use8BitMIME` (should fix)

`calendarInviteText` accepts any valid UTF-8 and the part is unconditionally labeled `Content-Transfer-Encoding: 7bit`. Non-ASCII is the norm for real invites (SUMMARY/LOCATION/attendee CN — "Besprechung mit Herrn Müller"). I verified with a scratch test on this branch: with `use8BitMIME: false` (server did not advertise 8BITMIME), the built message contains raw 8-bit octets in a part labeled 7bit — while the text body of the *same message* is dutifully quoted-printable encoded by `MIMEBuildContext`. So the change quietly breaks the invariant the rest of `constructContent` maintains: nothing 8-bit goes on the wire unless the server negotiated it. On a 7-bit-only path an MTA may strip high bits, corrupting the invite; and `prepareEmailForSend` has no post-hoc guard (only `sendRawMessage` does).

This also violates RFC 2045 §6.2 (7bit means no octets > 127) — RFC 6047's examples are 7bit because they're pure ASCII, not because the label is free.

Since the stated reason to avoid base64/QP is that *re-encoding* breaks client recognition, an honest `8bit` label keeps the bytes identical on the wire and costs nothing for recognition. Suggested shape (the flag is already in scope in `writeBody`; `writeMultipartMixed` would need it threaded through, e.g. via `MIMEBuildContext`):

```swift
// returns nil → regular base64 attachment path
private func calendarInviteText(for attachment: Attachment, use8BitMIME: Bool) -> (text: String, encoding: String)? {
    // … mime check, UTF-8 decode, CRLF normalization as today …
    guard icsText.isSafe8BitContent() else { return nil }          // NUL/control chars, 998-char lines
    if icsText.utf8.allSatisfy({ $0 < 128 }) { return (icsText, "7bit") }
    if use8BitMIME { return (icsText, "8bit") }
    return nil                                                     // non-ASCII on a 7-bit server
}
```

Falling back to base64 for non-ASCII on non-8BITMIME servers is a strict improvement over today's behavior (those clients were equally broken pre-PR), and in practice nearly every modern server advertises 8BITMIME, so the field-tested path is unchanged.

### Finding 2 — no line-length guard on the verbatim path (should fix)

RFC 5545 requires folding at 75 octets, but plenty of real-world ICS producers don't fold (long `DESCRIPTION`, `X-ALT-DESC` with HTML). Verified on this branch: an unfolded 1500-char `DESCRIPTION` produces a 1512-character line in the wire content, past the RFC 5321 998-octet limit — servers may reject or mangle the message. The old base64 path wrapped at 76 chars, so this failure mode is new.

The `isSafe8BitContent()` gate in the sketch above covers this (it checks the 998 limit plus NUL/control characters), reusing the exact mechanism the text/HTML bodies already go through.

### Minor

- **`Content-Disposition: inline` without a filename** (mixed path): the existing inline-attachment writer emits `inline; filename="…"`, and Gmail's own outgoing invites carry a filename too. Including `filename="\(attachment.filename)"` keeps the invite renderable while letting non-calendar clients save the .ics. Worth confirming against your field setup before changing, since you've validated the current shape.
- **`method=` parameter is the caller's job**: the tests always pass `text/calendar; method=REQUEST; charset=UTF-8`, but a bare `text/calendar` mimeType also takes this path, and clients (Outlook especially) won't show Accept/Decline without the `method` parameter. A doc note on the new behavior (or in an example) telling callers to include `; method=REQUEST` would prevent confusing "it didn't render" reports.
- **Tests to add with the fix**: a non-ASCII invite under both `use8BitMIME` values, and an over-998-char-line ICS falling back to base64.

Note: CI hasn't run on this PR yet (no checks reported — likely awaiting first-time-contributor approval).

odrobnik:
@hjbrandt would you mind addressing the review comments?

hjbrandt:
Thanks for the thorough review — both findings were real and are fixed (pushed in e515eab).

**Finding 1 (7bit label on non-ASCII).** `calendarInviteText` now returns `(text, encoding)` and labels honestly: pure-ASCII ICS → `7bit` (the field-proven shape, byte-identical to before); non-ASCII → `8bit` only when the server advertised 8BITMIME; non-ASCII without 8BITMIME → falls back to base64. `use8BitMIME` is threaded through `MIMEBuildContext` to both the `multipart/alternative` and `multipart/mixed` paths, so the all-ASCII field path is unchanged.

**Finding 2 (no line-length guard).** Added a 998-limit check on the verbatim path. One refinement over reusing `isSafe8BitContent()` directly: I measure UTF-8 **octets** between hard CRLFs rather than calling that helper for the length test — it counts grapheme clusters, so a 600-character non-ASCII `SUMMARY` (1200 octets) would slip through, and it splits on U+2028/U+2029 which SMTP does not treat as line breaks. Since non-ASCII is exactly what this PR newly routes through 8bit, the octet-accurate check is what actually enforces the limit. `isSafe8BitContent()` still gates NUL/control bytes. Over-long lines fall back to base64.

**Minors.** The inline calendar part now carries `Content-Disposition: inline; filename="…"` (matching Gmail's outgoing invites). Added a doc note that callers must include `; method=REQUEST` on the mimeType for Outlook to offer Accept/Decline.

**Tests.** Five cases added/updated: non-ASCII → 8bit under 8BITMIME, non-ASCII → base64 without it, over-long ASCII line → base64, over-long multibyte line → base64 (the octet-vs-grapheme case), and the mixed-path test now asserts the inline filename. Full suite green (297).

odrobnik:
@hjbrandt Please also do swift lint adjustments to fit the linting rules

hjbrandt:
Done — wrapped the 9 `line_length` violations in the iMIP tests (string values unchanged, suite still green). Pushed in d7c019e.

PR #179, opened 2026-06-23
Make the IMAP response parser buffer limit configurable

tabmail-kmyi:
## What

The IMAP response parser's `bufferLimit` is hardcoded to 1 MB in `IMAPConnection`'s connection bootstrap. This makes it configurable via a new `responseBufferLimit` parameter on `IMAPServer`, defaulting to 1 MB so existing callers are unaffected.

## Why

Dense mailboxes can return `SEARCH`/`FETCH` responses larger than 1 MB — e.g. a `UID SEARCH` matching tens of thousands of messages. These overflow the parser's buffer and surface as `PayloadTooLargeError`, which drops the connection. Today the only recourse is to live with the failures or fork the constant. Exposing it lets clients that index large mailboxes raise the limit when they need to.

## What changed

- Add `IMAPServer.defaultResponseBufferLimit` (1 MB) and a `responseBufferLimit:` parameter on both public `IMAPServer` initializers, defaulting to it — **no behavior change** for existing callers.
- Thread the value to the primary connection and to spawned **idle** and **named** connections (where large IDLE / parallel-FETCH responses actually arrive), then apply it to `ResponseParser.Options`.
- The connect log line now reports the actual configured limit instead of a hardcoded "1MB".

## Tests

Adds `IMAPResponseBufferLimitTests`:
- default resolves to `defaultResponseBufferLimit` (1 MB),
- a custom value propagates through both initializers to the primary connection,
- the `IMAPConnection`-level plumbing honors default and custom limits.

Full suite green locally (310 tests, Xcode 26 / Swift 6.3.2).

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
@@ -28,7 +28,7 @@ extension IMAPConnection {
         self.isSessionAuthenticated = false
         self.namespaces = nil
 
-        logger.info("\(connectionContext) Connected to IMAP server with 1MB buffer limit for large responses")
+        logger.info("\(connectionContext) Connected; response buffer limit \(responseBufferLimit) bytes")
 
         let greetingCapabilities = try await waitForGreeting(
             channel: channel,
@@ -47,6 +47,7 @@ extension IMAPConnection {
         let certificateVerificationPolicy = self.certificateVerificationPolicy
         let duplexLogger = self.duplexLogger
         let responseBuffer = self.responseBuffer
+        let responseBufferLimit = self.responseBufferLimit
 
         return ClientBootstrap(group: group)
             .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
@@ -54,7 +55,7 @@ extension IMAPConnection {
             .channelInitializer { channel in
                 do {
                     let parserOptions = ResponseParser.Options(
-                        bufferLimit: 1024 * 1024,
+                        bufferLimit: responseBufferLimit,
                         messageAttributeLimit: .max,
                         bodySizeLimit: .max,
                         literalSizeLimit: IMAPDefaults.literalSizeLimit
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -18,6 +18,7 @@ final class IMAPConnection {
     let port: Int
     let transportSecurity: MailTransportSecurity
     let certificateVerificationPolicy: MailCertificateVerificationPolicy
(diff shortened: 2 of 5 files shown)


odrobnik (commented):
I checked out the branch, reviewed the full diff against the surrounding code, built it, and ran the new tests.

Verdict: solid, ship-ready.

What changed looks correct: the IMAP response parser buffer limit is now configurable via defaulted initializer parameters, and the value is threaded through all three IMAPConnection construction sites without changing existing caller behavior.

What I verified:
- The plumbing is complete across the primary, idle, and named connection paths.
- The channel initializer captures a local responseBufferLimit value instead of self, matching the existing bootstrap pattern.
- The change is source-compatible because the new parameters are defaulted.
- The stale hardcoded 1 MB references were updated, including the connect log output.
- Build completed cleanly and the 5 new IMAPResponseBufferLimitTests passed.

Minor non-blocking notes:
1. There is no lower-bound validation for responseBufferLimit, so 0 or a negative value would effectively make every response fail with PayloadTooLargeError. A precondition or clamp could make that failure mode clearer.
2. The tests shut down their EventLoopGroup via fire-and-forget Task blocks from synchronous test code. It works, but syncShutdownGracefully() would be tidier and deterministic.
3. The two ForTesting accessors slightly widen the production API surface. That is acceptable here, but wrapping them in a debug-only condition would keep them out of release builds.

None of these block merge.

odrobnik:
@tabmail-kmyi if you could address the 3 minor notes, I could merge it

tabmail-kmyi:
Thanks for the careful review! Addressed all three:

**1. Lower-bound validation** — added `precondition(responseBufferLimit > 0, …)` to the public `IMAPServer` initializer (the `useTLS` convenience init forwards to it), plus a `- Precondition:` doc note. A non-positive limit now fails loudly at construction instead of silently making every response a `PayloadTooLargeError`.

**2. Deterministic group teardown** — done with `syncShutdownGracefully()` as you suggested. One wrinkle worth flagging: calling it *directly* from the test deadlocked the macOS CI job (a ~7-minute hang to the job timeout). swift-testing runs tests concurrently on the Swift Concurrency cooperative pool, and a blocking wait there violates the pool's forward-progress guarantee on a core-constrained runner (passes fine locally on many cores; NIO even asserts against `syncShutdownGracefully()` on an event loop for the same reason). So I keep `syncShutdownGracefully()` but dispatch it to a non-cooperative GCD global queue and `await` a continuation — the call is still `syncShutdownGracefully()` and fully deterministic (the test awaits its completion), it just never blocks a cooperative-pool thread. macOS CI is green again.

**3. Debug-only test accessors** — wrapped both `…ForTesting` accessors in `#if DEBUG` so they're stripped from release builds.

Verified locally (build + all tests + `swiftlint --strict`) and green on CI.

PR #180, opened 2026-06-29
Recycle IMAP IDLE connection without tagged completion

jdudley:
Summary:
- Treat an IDLE connection as reusable only after `DONE` reaches the final tagged IDLE completion.
- Recycle the connection when IDLE teardown is cancelled, when the channel/server terminates during teardown, or when completion is otherwise uncertain.
- Keep resilient IDLE session shutdown single-owner/idempotent and await the runner before shutting down the dedicated event-loop group.
- Prevent child tasks from inheriting `IMAPCommandQueue` ownership, so a child task cannot start another tagged command while its parent still owns the queue.
- Add regression coverage for cancelled `DONE`, untagged `BYE` during `DONE`, child-task queue reentry, and repeated IDLE renewals.

Review follow-up:
- Removed the earlier pipelined-fetch recycling change from this PR; this now stays focused on IDLE lifecycle/command serialization.
- Simplified the idle-session lifecycle so the stop path owns connection removal and event-loop shutdown.
- Await event-loop group teardown in the touched tests instead of launching detached cleanup tasks.
- Added a positive renewal assertion: the renewal test now verifies multiple IDLE cycles happened.
- Documented the lifetime invariant for the stored `UnsafeCurrentTask` references in `IMAPCommandQueue`.

Validation:
- `swift test --filter IMAPIdleCancellationTests` — 3 Swift Testing tests passed locally on the previous code head.
- `swift test --filter IMAPIdleGroupLifecycleTests` — 3 Swift Testing tests passed locally on the previous code head.
- `swift test --filter IdleHandlerTests` — 2 Swift Testing tests passed locally on the previous code head.
- `swift test --filter IMAPCommandQueueTests` — 3 Swift Testing tests passed locally after the comment-only follow-up.
- `swiftlint lint --strict --no-cache` — 0 violations locally after the comment-only follow-up.
- `swift test` — 315 Swift Testing tests passed locally on the previous code head.
- GitHub Actions were green for the previous head; checks are rerunning for head `c09ec24`.


odrobnik:
## Review

Thanks for this — the core direction is right. The cancelled-DONE handling in `performIdleDone` fixes a real pre-existing desync: before this PR a cancelled DONE rethrew while the `defer` still wiped `idleHandler`/`hasActiveHandler`, leaving a live channel that could still be mid-IDLE server-side with local state cleared. Recycling the transport there is correct. The new tests also genuinely drive the paths they target (I checked: `channel == nil` is only reachable via the new `disconnectBody()` on those paths, so they'd fail if the fix were reverted).

A few things worth addressing:

### 1. Pipelined fetch recycles a *healthy* authenticated connection on any per-part failure — `IMAPConnection+PipelinedFetch.swift:86`

```swift
if results.count != requests.count || !channel.isActive {
    try? await disconnectBody()
}
```

`results.count != requests.count` is the right signal for a **timeout** (responses still in flight → stream desynced) or a dead channel. But it also fires when a single part fails at the *application* level: `PipelinedFetchPartHandler.processTaggedResponse` fails that one part's future on any non-`.ok` tagged response (`IMAPError.fetchFailed`), and `collectPipelinedFetchResults` drops it. So a server returning a tagged `NO`/`BAD` for one UID (e.g. a UID expunged between SEARCH and FETCH, or a transient per-message error) — while the stream is fully synchronized and the channel is healthy — trips this and calls `disconnectBody()`, which wipes the channel **plus `isSessionAuthenticated`, `capabilities`, and `namespaces`**. The next operation then pays a full reconnect + re-auth + re-SELECT.

That contradicts the PR's own goal ("avoid returning an *uncertain* command state"): a tagged `NO` is a *certain*, completed state. Suggest distinguishing the two failure modes — recycle on timeout / `!channel.isActive` / transport+parse failure, but keep the connection when every command got a tagged response and only the *data* was unavailable. (The handler already knows the difference: a part resolved by a tagged response vs. one failed by the timeout/`channelInactive` path.)

### 2. `IMAPIdleSessionLifecycle.finishFromRunner()` shutdown is dead code — `IMAPServer+Idle.swift:184`

`IMAPResilientIdleRunner.run` is `while !Task.isCancelled`, and every error/BYE/timer path loops back (`handleCycleError` never rethrows). So the runner exits **only** on cancellation, and the only canceller — `beginStop` — sets `isStopped = true` first. By the time the detached task reaches `finishFromRunner()`, `guard !isStopped` is always true, so `idleGroup.shutdownGracefully()` inside it never runs (the real shutdown is in `finishStop`). The whole two-path `isStopped` machinery coordinates a race that can't happen with the current runner.

Either simplify (drop `finishFromRunner`, keep `continuation.finish()` in the task and do cancel → `await cycleTask.value` → `shutdownGracefully()` in the stop closure), or at minimum add a comment documenting the "runner exits only on cancellation" invariant. It's load-bearing: if a future change ever lets `run` return on its own, `beginStop` would return `false` on the user's `done()`, so `endIdleSession(id:)` would never run and the connection would leak in `idleConnections`.

### 3. Fire-and-forget cleanup in `.serialized` suites — `PipelinedFetchTests.swift:120`, `IMAPIdleCancellationTests.swift:24`

```swift
defer { Task { await harness.shutdown() } }   // group.shutdownGracefully() detached, never awaited
```

The `MultiThreadedEventLoopGroup` teardown is detached and not awaited, so it can still be running when the next serialized test sets up its own channel/group — nondeterministic leaked-ELG warnings / cross-test bleed. Swift Testing async tests support `await` directly in `defer`; just `await harness.shutdown()`.

### 4. `repeatedIdleRenewalsDoNotTripClientStateMachine` has no positive assertion — `IMAPIdleGroupLifecycleTests.swift:75`

The test has no `#expect`/`Issue.record`; it only fails if the final `try await session.done()` throws. It would still pass if renewals never actually fired (mis-wired interval + a single clean `done()` succeeds). Consider asserting that cycles ran / the session is still healthy after the sleep.

### Minor

- The new `disconnectBody()` calls (here and the dispatch-error catch) hand-roll recycling rather than going through the established `shouldRecycleConnection(for:)` / `clearInvalidChannel()` path the other command sites use. Mostly defensible, but worth keeping the recycle policy in one place — `shouldRecycleConnection` deliberately does *not* recycle on `CancellationError`, and the divergence is easy to lose track of.

For what it's worth, I checked the concurrency-sensitive parts of the lifecycle rewrite (the `finishStop`/`finishFromRunner` reentrancy on the same actor, double `shutdownGracefully`, concurrent/repeated `done()`, the `[weak self]` nil branch, and the cancel-vs-disconnect ordering) and they're all safe — `done()` is now nicely idempotent via the `isStopped` gate.

jdudley:
Follow-up from review on current head `8fa8504`:

- Updated the PR title/body to frame the invariant: an IDLE connection is reusable only after the final tagged IDLE completion.
- Removed the earlier pipelined-fetch recycling change from this PR; the remaining diff is IDLE lifecycle plus command serialization.
- Simplified the IDLE session lifecycle so the stop path owns connection removal and event-loop shutdown.
- Updated touched tests to await event-loop group teardown.
- Added/kept positive coverage for cancelled `DONE`, untagged `BYE` during `DONE`, child-task command queue reentry, and repeated IDLE renewals.

Validation on this branch:
- `swift test --filter IMAPIdleCancellationTests` passed 3 tests.
- `swift test --filter IMAPIdleGroupLifecycleTests` passed 3 tests.
- `swift test --filter IMAPCommandQueueTests` passed 3 tests.
- `swift test --filter IdleHandlerTests` passed 2 tests.
- `swiftlint lint --strict --no-cache` found 0 violations.
- `swift test` passed 315 tests.
- GitHub Actions are green for macOS, iOS, Linux, Android, and SwiftLint; Windows is skipped by workflow configuration.

odrobnik:
## Re-review of `8fa8504`

Re-reviewed the three new commits. All earlier points are resolved, and the two new pieces hold up:

**Prior findings — all addressed**
1. Pipelined-fetch recycle-on-partial — removed entirely (along with its test). 👍 keeps the PR scoped to IDLE.
2. `finishFromRunner` dead code / over-built actor — gone; the actor is now just the `isStopped` gate and the stop closure owns `await cycleTask.value` → `shutdownGracefully()`, with the cancellation-only-exit invariant documented in the comment.
3. Fire-and-forget test cleanup — now awaited via the `do { … try await group.shutdownGracefully() } catch { … }` wrapper (the right call, since `defer` can't `await`).
4. `repeatedIdleRenewals…` — now asserts `testServer.idleCommandCount >= 2`, so it actually proves renewals fired.

**New: untagged-termination recycling (`requireTaggedIdleCompletion`) — correct.** Keying the recycle decision on `IdleHandler.completionReason` is the right altitude — it cleanly separates "DONE got its tagged completion" (reuse) from BYE / channelInactive / error (recycle), per RFC 2177's requirement that the tagged response retires the IDLE tag. The reason is recorded first-wins per handler and each cycle gets a fresh handler, so there's no staleness across renewals. The cancellation path still short-circuits ahead of it, and `endIdleSession` swallows the new throw via `try?`, so the resilient runner just reconnects on a BYE-during-DONE. Behavior change worth calling out in the changelog: a direct `connection.done()` now throws `connectionFailed` (instead of returning) when the server ends IDLE with an untagged BYE — intended, but a caller-visible difference.

**New: command-queue task-identity ownership — correct, one note.** Switching ownership from an inherited `@TaskLocal` UUID to the current task's identity genuinely fixes the child-task reentry hole (an unstructured `Task {}` inherited the parent's token and could interleave). Nicely covered by `testChildTaskDoesNotInheritQueueOwnership` and `childTaskIdleStartWaitsForInFlightTaggedCommand`.

One thing worth a code comment (not a blocker): `UnsafeCurrentTask` is documented as *"Don't store an unsafe task reference for later use"*, and the queue stores it in `ownerTask` / `Waiter.ownerTask` and compares it later. It's safe *here* because the stored owner is always a task currently inside `run()` — either executing `operation()` or suspended in `acquire`'s continuation — so it can't be deallocated during an `isSameTask` comparison, which rules out a dangling-pointer / address-reuse false match. That's a subtle invariant to be relying on silently. Suggest either documenting it at the `ownerTask` declaration, or — since the codebase already avoids nested same-task `run` (see the "would deadlock" note in `applyXOAUTH2Capabilities`) — considering whether same-task reentrancy is needed at all; if not, a plain non-reentrant FIFO drops the task-identity tracking entirely.

LGTM with that as an optional follow-up. Thanks for the thorough test coverage and the green cross-platform run.

jdudley:
Addressed the optional `UnsafeCurrentTask` documentation note in `c09ec24`:

- Added comments documenting that stored task references are only held while the owning task is inside `run()` or while a waiter is suspended in `acquire`.
- No behavior change.

Validation after the comment-only change:
- `swift test --filter IMAPCommandQueueTests` passed 3 tests.
- `swiftlint lint --strict --no-cache` found 0 violations.

PR #181, opened 2026-06-29
Coalesce concurrent named IMAP connection authentication

jdudley:
Summary:
- Coalesce concurrent `IMAPServer.connection(named:)` creation for the same normalized connection name.
- Coalesce concurrent authentication attempts inside `IMAPNamedConnection` so same-handle callers share one in-flight authentication attempt.
- Preserve retry behavior after failures by clearing the in-flight sentinel before returning to later callers.
- Add regression coverage for same-name connection sharing and same-handle authentication sharing.

Review follow-up:
- Await event-loop group shutdown in the touched tests instead of launching detached teardown tasks.
- Replace the fixed-sleep auth regression check with a deterministic waiter-count signal.
- Keep the single-flight helper local to the two actor-owned call sites; this avoids introducing a broader abstraction for a private two-site pattern.
- Coalesced waiters intentionally inherit the leader result. If that result fails, a later call can retry normally.
- Make the internal waiter-count testing hook available in release test builds; it remains internal and only reachable from tests via `@testable import`.

Validation:
- `swift test --filter IMAPNamedConnectionTests` — 6 Swift Testing tests passed locally.
- `swiftlint lint --strict --no-cache` — 0 violations locally.
- `swift test -c release --filter IMAPNamedConnectionTests` — no longer reports the `authenticationWaiterCountForTesting` error; the release build still stops on existing `IMAPResponseBufferLimitTests` testing hooks outside this PR.
- GitHub Actions for the previous head were green for macOS, iOS, Linux, Android, and SwiftLint; checks are rerunning for head `a61413b`.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -41,6 +41,15 @@ jobs:
         uses: maxim-lobanov/setup-xcode@v1
         with:
           xcode-version: "26.0"
+      # Xcode 16+ ships platform SDKs as on-demand components, so a freshly
+      # rotated `macos-latest` image can have the selected Xcode without the iOS
+      # platform installed — `-destination 'generic/platform=iOS'` then resolves
+      # to zero destinations ("iOS … is not installed") before anything compiles.
+      # Fetch the component when it is missing. Best-effort (`|| true`): on images
+      # that already have it the download itself can error, and the Build step
+      # below is the real gate either way (it only needs the SDK).
+      - name: Ensure the iOS platform is installed (best-effort)
+        run: xcodebuild -downloadPlatform iOS || true
       - name: Build (iOS)
         run: xcodebuild -scheme SwiftMail -destination 'generic/platform=iOS' build
 
--- a/Sources/SwiftMail/IMAP/IMAPNamedConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPNamedConnection.swift
@@ -21,11 +21,9 @@ public actor IMAPNamedConnection {
     private var authenticationWaiters: [CheckedContinuation<Void, any Error>] = []
     private var isAuthenticationInFlight = false
 
-    #if DEBUG
     var authenticationWaiterCountForTesting: Int {
         authenticationWaiters.count
     }
-    #endif
 
     init(
         name: String,
--- a/Sources/SwiftMail/IMAP/IMAPNamedConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPNamedConnection.swift
@@ -18,6 +18,8 @@ public actor IMAPNamedConnection {
     /// The timestamp of the last successfully completed command on this connection.
     /// Useful for implementing staleness checks in ephemeral connection patterns.
     public private(set) var lastActivity: Date?
+    private var authenticationWaiters: [CheckedContinuation<Void, any Error>] = []
+    private var isAuthenticationInFlight = false
 
     init(
         name: String,
@@ -70,8 +72,37 @@ public actor IMAPNamedConnection {
     }
 
     func ensureAuthenticated() async throws {
-        if !connection.isAuthenticated {
+        guard !connection.isAuthenticated else { return }
+
+        if isAuthenticationInFlight {
+            try await withCheckedThrowingContinuation { continuation in
+                authenticationWaiters.append(continuation)
+            }
+            return
+        }
+
+        isAuthenticationInFlight = true
+        do {
             try await authenticateOnConnection(connection)
+            completeAuthenticationWaiters(with: .success(()))
+        } catch {
+            completeAuthenticationWaiters(with: .failure(error))
+            throw error
+        }
+    }
+
+    private func completeAuthenticationWaiters(with result: Result<Void, any Error>) {
+        isAuthenticationInFlight = false
+        let waiters = authenticationWaiters
+        authenticationWaiters.removeAll()
+
+        for waiter in waiters {
+            switch result {
+                case .success:
+                    waiter.resume()
+                case .failure(let error):
+                    waiter.resume(throwing: error)
+            }
         }
     }
 
--- a/Sources/SwiftMail/IMAP/IMAPNamedConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPNamedConnection.swift
@@ -21,6 +21,12 @@ public actor IMAPNamedConnection {
     private var authenticationWaiters: [CheckedContinuation<Void, any Error>] = []
     private var isAuthenticationInFlight = false
 
+    #if DEBUG
+    var authenticationWaiterCountForTesting: Int {
+        authenticationWaiters.count
+    }
+    #endif
+
     init(
         name: String,
         connection: IMAPConnection,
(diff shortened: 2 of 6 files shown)


odrobnik:
## Review — single-flight coalescing for named connections + auth

The core coalescing is solid. I applied the diff to the base and traced every suspension point in both actors:

- **No duplicate-leader race** — the existing-connection lookup, the `pendingNamedConnectionWaiters[name]` check, and the `… = []` sentinel set are all synchronous within the actor (no `await` between them), so two callers can't both become leader. Same for `isAuthenticationInFlight` in `ensureAuthenticated`.
- **No leaked continuations** — every sentinel-set path has a matching `complete…` call on both the success and the `catch` path, and waiters are appended synchronously before suspending, so none can be parked forever.
- **No double-resume** — each `complete…` reads-then-clears its batch and resumes each continuation exactly once.

I also confirmed both new tests genuinely fail without the fix (real regressions, not tautologies): base `ensureAuthenticated` has no guard, so two callers both invoke the auth closure → count 2; base `connection(named:)` only populates the cache *after* the `connect()`/`authenticate()` `await`, so two concurrent callers get two distinct handles → the `ObjectIdentifier` assertion fails.

A few things worth considering, none blocking:

### Behavior — coalesced callers inherit the leader's outcome (confirm intended)

With single-flight, if the leader's `connect()`/`authenticate()` fails, every parked waiter is resumed with the *leader's* error and gets no attempt of its own — whereas pre-PR each concurrent caller attempted independently. In practice coalesced callers share the same host/credentials, so they'd usually share the outcome anyway; the case that changes is a transient leader-only failure (e.g. the 5 s greeting timeout firing on a momentary blip) now taking the waiters down with it.

Related: waiters park on a bare `withCheckedThrowingContinuation` with no cancellation handling, so a waiter task cancelled while parked isn't released until the leader completes. (To be clear, this is *not* a `CancellationError`-poisoning issue — the connect/auth path awaits NIO `EventLoopFuture.get()`, which doesn't observe Swift task cancellation, so a cancelled leader runs to completion rather than throwing into the waiters.) If you'd want cancelled waiters to abort promptly, or a healthy waiter to re-lead on leader failure, that needs `withTaskCancellationHandler` around the park and/or re-promoting a waiter — fine to leave as the intended single-flight tradeoff, just flagging it.

### Cleanup — the single-flight mechanism is written twice

`completePendingNamedConnection(named:with:)` (`IMAPServer+Connection.swift`) and `completeAuthenticationWaiters(with:)` (`IMAPNamedConnection.swift`) are the same pattern — an array of `CheckedContinuation` + a `switch` that resumes every waiter from a `Result` — and the two sites even encode "a leader is in flight" two different ways: a dedicated `isAuthenticationInFlight: Bool` in one, key-presence (`pendingNamedConnectionWaiters[name] != nil`, seeded with `[]`) in the other. A small generic actor coalescer would collapse both call sites and keep the two copies of the continuation-resume logic from drifting. (No existing helper fits — `IMAPCommandQueue` is a serial mutex, not a value-sharing coalescer.)

Minor, same area: in the waiter branch, `pendingNamedConnectionWaiters[normalizedName, default: []].append(…)` is only reached after `… != nil` was just checked true, so the `default: []` can never fire — a plain subscript states the actual invariant.

### Tests

- **`makeConnection()` leaks its event-loop group** — it allocates `MultiThreadedEventLoopGroup(numberOfThreads: 1)` and returns with no shutdown path, unlike the sibling tests that `defer { Task { try? await group.shutdownGracefully() } }`. The new auth-coalescing test adds another leaked ELG thread for the process lifetime.
- **`concurrentAuthenticationRequestsShareSingleInFlightAttempt` leans on a fixed 50 ms sleep** — the busy-wait only proves the *first* task started; the assertion then sleeps 50 ms and checks `startCount() == 1`. On a slow runner the *second* task might not reach `startAndWait()` within the window, so it can pass even if coalescing regressed (a weakened regression, not a flaky failure). Signalling from inside the second task would make it deterministic.

(The `#if os(macOS)` gate on the connection test matches the existing convention for every `IMAPTestServer`-based test, and the auth-coalescing test already runs cross-platform — so no concern there.)

jdudley:
Follow-up from review on current head `0ed9111`:

- Updated the PR body with current scope and validation.
- Awaited event-loop group shutdown in the touched tests.
- Replaced the fixed-sleep auth regression check with a deterministic waiter-count signal.
- Kept the single-flight implementation local to the two private actor-owned call sites instead of introducing a generic helper for this PR.
- Documented the intended tradeoff: coalesced waiters inherit the leader result, and later calls can retry normally after a failure.

Validation on this branch:
- `swift test --filter IMAPNamedConnectionTests` passed 6 tests.
- `swiftlint lint --strict --no-cache` found 0 violations.
- GitHub Actions are green for macOS, iOS, Linux, Android, and SwiftLint; Windows is skipped by workflow configuration.

odrobnik:
## Re-review (commit `0ed9111`)

Thanks for the quick turnaround — the feedback's been addressed well:

- ✅ **ELG leak** — `NamedConnectionHarness` + `shutdown()` now cleans up the group at every call site, and you tightened the two sibling tests (`defer { Task { … } }` fire-and-forget → an awaited `shutdownGracefully()`) while you were at it.
- ✅ **Weak 50 ms sleep** — `waitForAuthenticationWaiter` now polls the actual waiter count and fast-fails via `Issue.record` if a second auth ever starts, so the regression is deterministic instead of wall-clock-dependent.
- ✅ **Dead `default: []`** — replaced with `pendingNamedConnectionWaiters[normalizedName]?.append(…)`.
- ✅ **Leader-coupling behavior** — documented as intended ("failure clears … so a later call can retry normally"). That was a confirm-intent note, not a defect, so 👍.

One new thing the fix introduces:

### `authenticationWaiterCountForTesting` is `#if DEBUG`, but the test references it unconditionally

`waitForAuthenticationWaiter(on:gate:)` reads `await named.authenticationWaiterCountForTesting`, and the property is compiled only in debug builds — so a release-configuration test build (`swift test -c release`) fails to compile the test target. It's also the only `#if DEBUG` in `Sources/`, and the only `*ForTesting` hook that's gated at all — every other one (`responseBufferLimitForTesting`, `replaceCapabilitiesForTesting`, `certificateVerificationPolicyForTesting`, …) is plain `internal`.

Simplest fix: drop the `#if DEBUG`, matching the existing convention. The property is already `internal`, so it stays out of the public API and is only reachable via `@testable import` regardless.

### Still optional (unchanged from last round)

The two single-flight fan-outs (`completeAuthenticationWaiters` / `completePendingNamedConnection`) remain separate implementations that encode "a leader is in flight" two different ways — a `Bool` in one, dict-key presence in the other. A small shared coalescer would dedupe them, but it's a nice-to-have, not a blocker.

jdudley:
Addressed in `a61413b`:

- Removed the `#if DEBUG` gate around `authenticationWaiterCountForTesting`.
- The helper remains `internal`, so it is still outside public API and only visible to tests through `@testable import`.

Validation:
- `swift test --filter IMAPNamedConnectionTests` passed 6 tests.
- `swiftlint lint --strict --no-cache` found 0 violations.
- `swift test -c release --filter IMAPNamedConnectionTests` no longer reports `authenticationWaiterCountForTesting`; it now stops on existing `IMAPResponseBufferLimitTests` release-build test hooks that are outside this PR.

PR #185, opened 2026-07-16
Add configurable minimum TLS version and IMAP parser limits

davyd15:
Two independent security hardening knobs, found while building a mail client on SwiftMail. Both were verified end-to-end against a real Hetzner IMAP/SMTP server (Dovecot, TLS 1.3).

They're separate commits and can be split or cherry-picked — happy to open two PRs instead if you prefer.

---

### 1. Minimum TLS version is not settable (`59cd258`)

`MailTLSConfiguration.makeClientConfiguration` builds on `TLSConfiguration.makeClientConfiguration()` and only overrides `certificateVerification` and `trustRoots`. It therefore inherits NIOSSL's default `minimumTLSVersion` of `.tlsv1` — TLS 1.0, [deprecated by RFC 8996](https://datatracker.ietf.org/doc/html/rfc8996) since March 2021. A caller has no way to raise that floor.

This adds `MailTLSMinimumVersion` and threads it through `IMAPServer`, `SMTPServer`, `IMAPConnection` and **both** SMTP TLS paths — the STARTTLS path at `SMTPServer+Connection.swift:283` builds its own config and would otherwise have stayed on the old default.

The enum is deliberately SwiftMail's own type rather than NIOSSL's `TLSVersion`: SwiftMail imports NIOSSL without re-exporting it, so exposing `TLSVersion` publicly would push NIOSSL onto every consumer.

**Behaviour change:** the default floor moves from TLS 1.0 to TLS 1.2, the lowest version RFC 8996 still permits. Callers needing the old behaviour can pass `.tlsv1`. **If you'd rather keep this strictly additive, say the word and I'll default it to `.tlsv1`.**

```swift
// Servers you know speak TLS 1.3 — downgrade becomes impossible
let server = IMAPServer(host: "imap.example.com", port: 993, minimumTLSVersion: .tlsv1_3)
```

### 2. Parser body/attribute limits are hardcoded to `.max` (`76d7a75`)

`IMAPConnection.makeBootstrap` pins the response parser to `messageAttributeLimit: .max` and `bodySizeLimit: .max`, with no way for a caller to bound them.

A server controls both the length prefixes and the attribute count it sends, so a hostile or malfunctioning server can force unbounded allocation from a single response. `responseBufferLimit` doesn't cover this — it bounds the parser's working buffer, whereas these bound what a single response may *declare*.

This is the class of issue that produced advisories in other IMAP clients: Ruby net-imap [GHSA-j3g3-5qv5-52mj](https://github.com/ruby/net-imap/security/advisories/GHSA-j3g3-5qv5-52mj) (unbounded literal size → memory exhaustion, fixed by adding a configurable `max_response_size`) and mutt [CVE-2021-3657](https://nvd.nist.gov/vuln/detail/CVE-2021-3657) ("reject excessively large IMAP literals").

Adds `IMAPParserLimits`. **Strictly additive** — defaults are `.max`/`.max`, exactly today's behaviour, so existing callers see no change.

```swift
let server = IMAPServer(
    host: "imap.example.com",
    port: 993,
    parserLimits: IMAPParserLimits(bodySizeLimit: 64 * 1024 * 1024, messageAttributeLimit: 1024)
)
```

Bounded defaults would arguably be the better long-term choice, but that would silently break anyone fetching large attachments — left that call to you.

Credit where due: the recursion depth **is** already bounded (`StackTracker(maximumParserStackDepth: 100)`), so the deeply-nested-MIME stack overflow that hit [Stalwart](https://github.com/stalwartlabs/stalwart/security/advisories/GHSA-jm95-876q-c9gw) and Dovecot (CVE-2020-12100) doesn't apply here.

---

### Testing

- Full suite passes: **331 tests in 42 suites**
- New: `MailTLSConfigurationTests` (default floor, each enum mapping, both server types) and `IMAPParserLimitsTests` (defaults, partial overrides, propagation)
- Verified against a live Hetzner server: with `minimumTLSVersion: .tlsv1_3` enforced, connect + login + CAPABILITY all succeed

Thanks for SwiftMail — it was the only actively maintained Swift IMAP client we could build on after MailCore2 went quiet.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
@@ -45,6 +45,7 @@ extension IMAPConnection {
     ) -> ClientBootstrap {
         let host = self.host
         let certificateVerificationPolicy = self.certificateVerificationPolicy
+        let minimumTLSVersion = self.minimumTLSVersion
         let duplexLogger = self.duplexLogger
         let responseBuffer = self.responseBuffer
         let responseBufferLimit = self.responseBufferLimit
@@ -55,18 +56,14 @@ extension IMAPConnection {
             .channelOption(ChannelOptions.tcpOption(.tcp_nodelay), value: 1)
             .channelInitializer { channel in
                 do {
-                    let parserOptions = ResponseParser.Options(
-                        bufferLimit: responseBufferLimit,
-                        messageAttributeLimit: parserLimits.messageAttributeLimit,
-                        bodySizeLimit: parserLimits.bodySizeLimit,
-                        literalSizeLimit: parserLimits.literalSizeLimit
-                    )
+                    let parserOptions = parserLimits.makeParserOptions(bufferLimit: responseBufferLimit)
 
                     if case .implicitTLS = initialTLSMode {
                         let sslHandler = try Self.makeTLSHandler(
                             for: channel,
                             host: host,
-                            certificateVerificationPolicy: certificateVerificationPolicy
+                            certificateVerificationPolicy: certificateVerificationPolicy,
+                            minimumTLSVersion: minimumTLSVersion
                         )
                         try channel.pipeline.syncOperations.addHandler(sslHandler)
                     }
--- a/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
@@ -50,6 +50,8 @@ extension IMAPConnection {
         let responseBuffer = self.responseBuffer
         let responseBufferLimit = self.responseBufferLimit
         let parserLimits = self.parserLimits
+        let logger = self.logger
+        let connectionContext = self.connectionContext
 
         return ClientBootstrap(group: group)
             .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
@@ -70,6 +72,16 @@ extension IMAPConnection {
 
                     try channel.pipeline.syncOperations.addHandlers([
                         IMAPClientHandler(parserOptions: parserOptions),
+                        // Directly behind the decoder: it sees every response and every
+                        // parser-limit error before any command handler does. The parser bounds
+                        // a single body section; this bounds the whole FETCH response, which is
+                        // what `bodySizeLimit` promises — and it closes the connection on a
+                        // violation instead of letting a rejected response sit in the decoder.
+                        IMAPResponseLimitGuard(
+                            bodySizeLimit: parserLimits.bodySizeLimit,
+                            logger: logger,
+                            connectionContext: connectionContext
+                        ),
                         duplexLogger,
                         greetingHandler,
                         responseBuffer
--- a/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
@@ -48,6 +48,7 @@ extension IMAPConnection {
         let duplexLogger = self.duplexLogger
         let responseBuffer = self.responseBuffer
         let responseBufferLimit = self.responseBufferLimit
+        let parserLimits = self.parserLimits
 
         return ClientBootstrap(group: group)
             .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
@@ -56,9 +57,9 @@ extension IMAPConnection {
                 do {
                     let parserOptions = ResponseParser.Options(
                         bufferLimit: responseBufferLimit,
-                        messageAttributeLimit: .max,
-                        bodySizeLimit: .max,
-                        literalSizeLimit: IMAPDefaults.literalSizeLimit
+                        messageAttributeLimit: parserLimits.messageAttributeLimit,
+                        bodySizeLimit: parserLimits.bodySizeLimit,
+                        literalSizeLimit: parserLimits.literalSizeLimit
                     )
 
                     if case .implicitTLS = initialTLSMode {
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -18,6 +18,7 @@ final class IMAPConnection {
     let port: Int
     let transportSecurity: MailTransportSecurity
     let certificateVerificationPolicy: MailCertificateVerificationPolicy
+    let minimumTLSVersion: MailTLSMinimumVersion
     let responseBufferLimit: Int
     let group: EventLoopGroup
     let connectionID: String
@@ -42,6 +43,7 @@ final class IMAPConnection {
         port: Int,
         transportSecurity: MailTransportSecurity = .automatic,
         certificateVerificationPolicy: MailCertificateVerificationPolicy = .fullVerification,
+        minimumTLSVersion: MailTLSMinimumVersion = .tlsv1_2,
         group: EventLoopGroup,
         loggerLabel: String,
         outboundLabel: String,
@@ -54,6 +56,7 @@ final class IMAPConnection {
         self.port = port
         self.transportSecurity = transportSecurity
         self.certificateVerificationPolicy = certificateVerificationPolicy
+        self.minimumTLSVersion = minimumTLSVersion
         self.responseBufferLimit = responseBufferLimit
         self.group = group
         self.connectionID = connectionID
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -39,12 +39,19 @@ final class IMAPConnection {
     let logger: Logging.Logger
     let duplexLogger: IMAPLogger
 
+    /// - Note: `minimumTLSVersion` and `parserLimits` deliberately have **no defaults.**
+    ///   They used to, and `makeIdleConnection`/`makeNamedConnection` simply left them out — so
+    ///   a server configured with a TLS 1.3 floor and 64 MiB parser limits spawned IDLE and named
+    ///   connections with a TLS 1.2 floor and *unbounded* limits, while `IMAPServer`'s
+    ///   documentation claimed the policy governed every transport. A default turns "forgot to
+    ///   pass the security policy" into "chose the lax one", and neither the compiler nor a
+    ///   reviewer can see the difference. Without defaults, omitting one is a build error.
     init(
         host: String,
         port: Int,
         transportSecurity: MailTransportSecurity = .automatic,
         certificateVerificationPolicy: MailCertificateVerificationPolicy = .fullVerification,
-        minimumTLSVersion: MailTLSMinimumVersion = .tlsv1_2,
+        minimumTLSVersion: MailTLSMinimumVersion,
         group: EventLoopGroup,
         loggerLabel: String,
         outboundLabel: String,
@@ -52,7 +59,7 @@ final class IMAPConnection {
         connectionID: String,
         connectionRole: String,
         responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit,
-        parserLimits: IMAPParserLimits = .default
+        parserLimits: IMAPParserLimits
     ) {
         self.host = host
         self.port = port
@@ -91,6 +98,11 @@ final class IMAPConnection {
         )
     }
 
+    /// Legacy convenience initializer taking `useTLS` instead of a `MailTransportSecurity`.
+    ///
+    /// Security policy is explicit here for the same reason the designated initializer has no
+    /// defaults: this call site was invisible to a grep for `IMAPConnection(` and only surfaced
+    /// once the compiler demanded the arguments.
     convenience init(
         host: String,
         port: Int,
@@ -101,20 +113,24 @@ final class IMAPConnection {
         inboundLabel: String,
         connectionID: String,
         connectionRole: String,
-        responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit
+        responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit,
+        minimumTLSVersion: MailTLSMinimumVersion = .tlsv1_2,
+        parserLimits: IMAPParserLimits = .default
     ) {
         self.init(
             host: host,
             port: port,
             transportSecurity: Self.resolveLegacyTransportSecurity(port: port, useTLS: useTLS),
             certificateVerificationPolicy: .fullVerification,
+            minimumTLSVersion: minimumTLSVersion,
             group: group,
             loggerLabel: loggerLabel,
             outboundLabel: outboundLabel,
             inboundLabel: inboundLabel,
             connectionID: connectionID,
             connectionRole: connectionRole,
-            responseBufferLimit: responseBufferLimit
+            responseBufferLimit: responseBufferLimit,
+            parserLimits: parserLimits
         )
     }
 
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -20,6 +20,7 @@ final class IMAPConnection {
     let certificateVerificationPolicy: MailCertificateVerificationPolicy
     let minimumTLSVersion: MailTLSMinimumVersion
     let responseBufferLimit: Int
+    let parserLimits: IMAPParserLimits
     let group: EventLoopGroup
     let connectionID: String
     let connectionRole: String
@@ -50,7 +51,8 @@ final class IMAPConnection {
         inboundLabel: String,
         connectionID: String,
         connectionRole: String,
-        responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit
+        responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit,
+        parserLimits: IMAPParserLimits = .default
     ) {
         self.host = host
         self.port = port
@@ -58,6 +60,7 @@ final class IMAPConnection {
         self.certificateVerificationPolicy = certificateVerificationPolicy
         self.minimumTLSVersion = minimumTLSVersion
         self.responseBufferLimit = responseBufferLimit
+        self.parserLimits = parserLimits
         self.group = group
         self.connectionID = connectionID
         self.connectionRole = connectionRole
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -114,7 +114,7 @@ final class IMAPConnection {
         connectionID: String,
         connectionRole: String,
         responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit,
-        minimumTLSVersion: MailTLSMinimumVersion = .tlsv1_2,
+        minimumTLSVersion: MailTLSMinimumVersion = .tlsv12,
         parserLimits: IMAPParserLimits = .default
     ) {
         self.init(
--- a/Sources/SwiftMail/IMAP/IMAPParserLimits.swift
+++ b/Sources/SwiftMail/IMAP/IMAPParserLimits.swift
@@ -39,7 +39,23 @@ public struct IMAPParserLimits: Sendable, Equatable {
     ///
     /// Message bodies are bounded by ``bodySizeLimit`` instead, so this can stay small.
     /// Defaults to `NIOIMAPCore.IMAPDefaults.literalSizeLimit` (4 KB).
+    ///
+    /// - Important: **Cannot exceed 8 KiB**, and the initializer rejects larger values rather
+    ///   than accept a limit it cannot keep. `IMAPClientHandler` constructs its
+    ///   `NIOSingleStepByteToMessageProcessor` with `maximumBufferSize:
+    ///   IMAPDefaults.lineLengthLimit` (8,192 bytes), hardcoded and independent of these
+    ///   options. A larger literal therefore parses only when the network happens to deliver it
+    ///   in few enough reads; if fragmentation leaves more than 8 KiB buffered before the
+    ///   literal completes, it fails with `PayloadTooLargeError` — below every configured limit.
+    ///   A configuration whose behaviour depends on packet boundaries is not a limit, it is a
+    ///   coin toss, so ``maximumSupportedLiteralSizeLimit`` is enforced up front.
     public var literalSizeLimit: Int
+
+    /// The largest value ``literalSizeLimit`` can take: `IMAPDefaults.lineLengthLimit` (8 KiB).
+    ///
+    /// Raising it further needs a configurable `maximumBufferSize` in `IMAPClientHandler`, which
+    /// the pinned NIOIMAP does not offer.
+    public static let maximumSupportedLiteralSizeLimit = IMAPDefaults.lineLengthLimit
 
     /// SwiftMail's previous behaviour: bodies and attribute counts unbounded.
     public static let `default` = IMAPParserLimits()
@@ -52,6 +68,15 @@ public struct IMAPParserLimits: Sendable, Equatable {
         precondition(bodySizeLimit > 0, "bodySizeLimit must be greater than 0")
         precondition(messageAttributeLimit > 0, "messageAttributeLimit must be greater than 0")
         precondition(literalSizeLimit > 0, "literalSizeLimit must be greater than 0")
+        precondition(
+            literalSizeLimit <= Self.maximumSupportedLiteralSizeLimit,
+            """
+            literalSizeLimit must not exceed \(Self.maximumSupportedLiteralSizeLimit) bytes: \
+            IMAPClientHandler caps its decoder buffer at IMAPDefaults.lineLengthLimit, so a \
+            larger literal only parses when the network happens to deliver it in few enough \
+            reads. Accepting the value would promise a bound that depends on packet boundaries.
+            """
+        )
         self.bodySizeLimit = bodySizeLimit
         self.messageAttributeLimit = messageAttributeLimit
         self.literalSizeLimit = literalSizeLimit
--- a/Sources/SwiftMail/IMAP/IMAPParserLimits.swift
+++ b/Sources/SwiftMail/IMAP/IMAPParserLimits.swift
@@ -56,4 +56,37 @@ public struct IMAPParserLimits: Sendable, Equatable {
         self.messageAttributeLimit = messageAttributeLimit
         self.literalSizeLimit = literalSizeLimit
     }
+
+    /// Translates these limits into `ResponseParser.Options`.
+    ///
+    /// ## Why this is a method and not four call sites
+    /// Because the translation is not the identity, and every place that spelled it out by hand
+    /// was a place it could drift. It already had: only the primary connection ever received
+    /// these limits at all.
+    ///
+    /// ## Why the `+ 1`
+    /// These properties are documented as *maxima* — a body of exactly `bodySizeLimit` is legal.
+    /// The parser, however, enforces strict inequalities:
+    ///
+    /// ```swift
+    /// guard size < self.bodySizeLimit else { throw ExceededMaximumBodySizeError(…) }
+    /// guard attributeCount < self.messageAttributeLimit else { throw … }
+    /// ```
+    ///
+    /// So passing the value through unchanged makes the effective maximum one *less* than the
+    /// documented one: `bodySizeLimit: 64 * 1024 * 1024` rejects a message of exactly 64 MiB, and
+    /// `messageAttributeLimit: 1` rejects a single-attribute FETCH — the attribute itself parses,
+    /// then the closing `)` runs the same guard with the count already incremented.
+    ///
+    /// Rather than redefine the public contract as an exclusive bound (surprising: nobody reads
+    /// "maximum" as "one less than this"), the conversion adds one. Saturating, because the
+    /// defaults are `.max` and overflow would trap — and there `.max` already means unbounded.
+    func makeParserOptions(bufferLimit: Int) -> ResponseParser.Options {
+        ResponseParser.Options(
+            bufferLimit: bufferLimit,
+            messageAttributeLimit: messageAttributeLimit == .max ? .max : messageAttributeLimit + 1,
+            bodySizeLimit: bodySizeLimit == .max ? .max : bodySizeLimit + 1,
+            literalSizeLimit: literalSizeLimit
+        )
+    }
 }
--- a/Sources/SwiftMail/IMAP/IMAPParserLimits.swift
+++ b/Sources/SwiftMail/IMAP/IMAPParserLimits.swift
@@ -1,0 +1,59 @@
+import NIOIMAPCore
+
+/// Bounds the IMAP response parser enforces against a malicious or malfunctioning server.
+///
+/// A server controls both the length prefixes and the number of attributes it sends. Without
+/// bounds, a hostile or broken server can force unbounded allocation from a single response —
+/// the same class of issue that produced advisories in other IMAP clients, for example Ruby's
+/// net-imap (GHSA-j3g3-5qv5-52mj, unbounded literal size) and mutt (CVE-2021-3657).
+///
+/// The defaults preserve SwiftMail's existing behaviour, so adopting this type changes nothing
+/// until a caller opts into tighter bounds:
+///
+/// ```swift
+/// let server = IMAPServer(
+///     host: "imap.example.com",
+///     port: 993,
+///     parserLimits: IMAPParserLimits(
+///         bodySizeLimit: 64 * 1024 * 1024,  // refuse absurd single messages
+///         messageAttributeLimit: 1024
+///     )
+/// )
+/// ```
+///
+/// - Note: ``IMAPServer/defaultResponseBufferLimit`` bounds the parser's *working buffer*.
+///   These limits bound what a single response may *declare*, which is a separate concern.
+public struct IMAPParserLimits: Sendable, Equatable {
+
+    /// Maximum size of message body data in a single response.
+    ///
+    /// Defaults to `UInt64.max` (unbounded), matching SwiftMail's previous behaviour.
+    public var bodySizeLimit: UInt64
+
+    /// Maximum number of attributes (`BODY`, `FLAGS`, `ENVELOPE`, …) in one FETCH response.
+    ///
+    /// Defaults to `Int.max` (unbounded), matching SwiftMail's previous behaviour.
+    public var messageAttributeLimit: Int
+
+    /// Maximum size of a single protocol literal — mailbox names and similar.
+    ///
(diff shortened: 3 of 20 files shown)


Sources/SwiftMail/IMAP/IMAPConnection.swift
```diff
@@ -42,19 +44,23 @@ final class IMAPConnection {
         port: Int,
         transportSecurity: MailTransportSecurity = .automatic,
         certificateVerificationPolicy: MailCertificateVerificationPolicy = .fullVerification,
+        minimumTLSVersion: MailTLSMinimumVersion = .tlsv1_2,
```
odrobnik: **HIGH** — Implicit-TLS IMAP connections never forward this value: `makeConnectionBootstrap` calls `makeTLSHandler` without `minimumTLSVersion`, so the handler silently uses its `.tlsv1_2` default. For example, `IMAPServer(port: 993, minimumTLSVersion: .tlsv1_3)` can still connect through a TLS-1.2-only endpoint, violating the caller's downgrade floor; choosing `.tlsv1` for a legacy TLS-1.0-only server likewise still fails because the actual floor remains 1.2.

Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -158,15 +171,19 @@ public actor IMAPServer {
...
-        responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit
+        responseBufferLimit: Int = IMAPServer.defaultResponseBufferLimit,
+        parserLimits: IMAPParserLimits = .default
     ) {
         precondition(responseBufferLimit > 0, "responseBufferLimit must be greater than 0 bytes")
         self.host = host
         self.port = port
         self.transportSecurity = transportSecurity
         self.certificateVerificationPolicy = certificateVerificationPolicy
+        self.minimumTLSVersion = minimumTLSVersion
```
odrobnik, line 184: **HIGH** — The configured TLS floor is passed only to `primaryConnection`; `makeIdleConnection` and `makeNamedConnection` omit it and construct connections with the `.tlsv1_2` default. A caller can successfully establish a TLS-1.3-only primary session and then have `connection(named:)` or `idle(on:)` negotiate TLS 1.2 through a downgraded endpoint, despite the server-level policy claiming to govern every transport.

Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -158,15 +171,19 @@ public actor IMAPServer {
...
         self.responseBufferLimit = responseBufferLimit
+        self.parserLimits = parserLimits
```
odrobnik, line 186: **HIGH** — The configured parser limits are passed only to `primaryConnection`; both spawned-connection factories omit `parserLimits` and therefore use `.default` (`.max` body and attribute limits). A hostile server can send an oversized or attribute-heavy FETCH response on a named or dedicated IDLE connection and bypass the caller's memory-exhaustion guards entirely.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
```diff
@@ -0,0 +1,59 @@
...
+/// ```
+///
+/// - Note: ``IMAPServer/defaultResponseBufferLimit`` bounds the parser's *working buffer*.
+///   These limits bound what a single response may *declare*, which is a separate concern.
+public struct IMAPParserLimits: Sendable, Equatable {
+
+    /// Maximum size of message body data in a single response.
+    ///
+    /// Defaults to `UInt64.max` (unbounded), matching SwiftMail's previous behaviour.
+    public var bodySizeLimit: UInt64
```
odrobnik, line 31: **MEDIUM** — This property is documented as an inclusive maximum, but the pinned NIOIMAP `ResponseParser` enforces `size < bodySizeLimit`. Consequently, a caller setting a 64 MiB maximum has a valid body of exactly 64 MiB rejected with `ExceededMaximumBodySizeError`, even though only data larger than the documented maximum should fail.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
```diff
@@ -0,0 +1,59 @@
...
+
+    /// Maximum number of attributes (`BODY`, `FLAGS`, `ENVELOPE`, …) in one FETCH response.
+    ///
+    /// Defaults to `Int.max` (unbounded), matching SwiftMail's previous behaviour.
+    public var messageAttributeLimit: Int
```
odrobnik, line 36: **MEDIUM** — The pinned parser checks `attributeCount < messageAttributeLimit` again when parsing the FETCH closing delimiter, making the effective maximum one less than this documented value. With `messageAttributeLimit: 1`, for example, a normal one-attribute `FETCH (FLAGS (...))` parses the attribute and then fails at `)`, so callers cannot successfully receive the advertised maximum count.

odrobnik (changes requested):
Requesting changes: the IMAP TLS floor is not honored for implicit TLS or spawned named/IDLE connections, spawned connections also bypass the configured parser limits, and the pinned response parser makes both advertised custom maxima effectively exclusive/off by one. SMTP's implicit and STARTTLS paths are consistently wired. Review was performed read-only; builds and tests were not run as requested.

Findings:

- **HIGH** `Sources/SwiftMail/IMAP/IMAPConnection.swift:47` — Implicit-TLS IMAP connections never forward this value: `makeConnectionBootstrap` calls `makeTLSHandler` without `minimumTLSVersion`, so the handler silently uses its `.tlsv1_2` default. For example, `IMAPServer(port: 993, minimumTLSVersion: .tlsv1_3)` can still connect through a TLS-1.2-only endpoint, violating the caller's downgrade floor; choosing `.tlsv1` for a legacy TLS-1.0-only server likewise still fails because the actual floor remains 1.2.
- **HIGH** `Sources/SwiftMail/IMAP/IMAPServer.swift:184` — The configured TLS floor is passed only to `primaryConnection`; `makeIdleConnection` and `makeNamedConnection` omit it and construct connections with the `.tlsv1_2` default. A caller can successfully establish a TLS-1.3-only primary session and then have `connection(named:)` or `idle(on:)` negotiate TLS 1.2 through a downgraded endpoint, despite the server-level policy claiming to govern every transport.
- **HIGH** `Sources/SwiftMail/IMAP/IMAPServer.swift:186` — The configured parser limits are passed only to `primaryConnection`; both spawned-connection factories omit `parserLimits` and therefore use `.default` (`.max` body and attribute limits). A hostile server can send an oversized or attribute-heavy FETCH response on a named or dedicated IDLE connection and bypass the caller's memory-exhaustion guards entirely.
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPParserLimits.swift:31` — This property is documented as an inclusive maximum, but the pinned NIOIMAP `ResponseParser` enforces `size < bodySizeLimit`. Consequently, a caller setting a 64 MiB maximum has a valid body of exactly 64 MiB rejected with `ExceededMaximumBodySizeError`, even though only data larger than the documented maximum should fail.
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPParserLimits.swift:36` — The pinned parser checks `attributeCount < messageAttributeLimit` again when parsing the FETCH closing delimiter, making the effective maximum one less than this documented value. With `messageAttributeLimit: 1`, for example, a normal one-attribute `FETCH (FLAGS (...))` parses the attribute and then fails at `)`, so callers cannot successfully receive the advertised maximum count.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
```diff
@@ -0,0 +1,92 @@
...
+/// ```
+///
+/// - Note: ``IMAPServer/defaultResponseBufferLimit`` bounds the parser's *working buffer*.
+///   These limits bound what a single response may *declare*, which is a separate concern.
+public struct IMAPParserLimits: Sendable, Equatable {
+
+    /// Maximum size of message body data in a single response.
+    ///
+    /// Defaults to `UInt64.max` (unbounded), matching SwiftMail's previous behaviour.
+    public var bodySizeLimit: UInt64
```
odrobnik, line 31: **HIGH** — `bodySizeLimit` is documented as the maximum body data in one response, but the pinned ResponseParser calls its size guard independently for every streaming BODY attribute and never tracks a response-wide total. `FetchPartHandler` appends every streaming attribute until the FETCH finishes. With `IMAPParserLimits(bodySizeLimit: 64 * 1024 * 1024)` and the default unbounded attribute count, a hostile server can return arbitrarily many 50 MiB BODY sections in one FETCH; every section passes the 64 MiB check while SwiftMail accumulates all of them into one `Data`, so the advertised memory-exhaustion guard remains bypassable.

Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
```diff
@@ -45,27 +45,25 @@ extension IMAPConnection {
...
             .channelOption(ChannelOptions.tcpOption(.tcp_nodelay), value: 1)
             .channelInitializer { channel in
                 do {
-                    let parserOptions = ResponseParser.Options(
-                        bufferLimit: responseBufferLimit,
-                        messageAttributeLimit: .max,
-                        bodySizeLimit: .max,
-                        literalSizeLimit: IMAPDefaults.literalSizeLimit
-                    )
+                    let parserOptions = parserLimits.makeParserOptions(bufferLimit: responseBufferLimit)
```
odrobnik, line 61: **HIGH** — Parser-limit errors are installed on IDLE connections, but they do not fail closed. During IDLE, an oversized BODY or over-limit attribute count makes `IMAPClientHandler` fire an error; `IdleHandler.errorCaught` only fails its private promise, without finishing the `AsyncStream` or closing the channel, and no task is awaiting that promise until a later DONE/checkpoint. The decoder retains the rejected response and each subsequent network read is appended before the same parse error is raised again (the post-decode buffer cap is skipped on throws), allowing a malicious peer to keep growing memory while the caller's IDLE stream hangs.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
```diff
@@ -0,0 +1,92 @@
...
+    /// Maximum number of attributes (`BODY`, `FLAGS`, `ENVELOPE`, …) in one FETCH response.
+    ///
+    /// Defaults to `Int.max` (unbounded), matching SwiftMail's previous behaviour.
+    public var messageAttributeLimit: Int
+
+    /// Maximum size of a single protocol literal — mailbox names and similar.
+    ///
+    /// Message bodies are bounded by ``bodySizeLimit`` instead, so this can stay small.
+    /// Defaults to `NIOIMAPCore.IMAPDefaults.literalSizeLimit` (4 KB).
+    public var literalSizeLimit: Int
```
odrobnik, line 52: **MEDIUM** — `literalSizeLimit` is exposed as a configurable maximum, but the pinned `IMAPClientHandler` constructs its byte-to-message processor with `maximumBufferSize: IMAPDefaults.lineLengthLimit` (8,192 bytes), regardless of these parser options or SwiftMail's 1 MiB response buffer limit. For example, a valid 16 KiB mailbox-name literal with `literalSizeLimit: 32 * 1024` succeeds only if delivered complete enough to parse immediately; if network fragmentation leaves more than 8 KiB buffered before the literal completes, it throws `PayloadTooLargeError` even though it is below every configured limit. Thus raised literal limits are not reliably honored.

odrobnik (changes requested):
Requesting changes. The five findings from the previous review are fixed at this head, but three blockers remain: `bodySizeLimit` is enforced per BODY attribute rather than across the single FETCH response that SwiftMail documents and may accumulate; an IDLE parser-limit violation leaves the socket and `AsyncStream` open while the decoder retains rejected bytes; and custom protocol-literal limits above the pinned handler's fixed 8 KiB aggregate cap fail depending on network fragmentation. Review was performed read-only; builds and tests were not run as requested.

Findings:

- **HIGH** `Sources/SwiftMail/IMAP/IMAPParserLimits.swift:31` — `bodySizeLimit` is documented as the maximum body data in one response, but the pinned ResponseParser calls its size guard independently for every streaming BODY attribute and never tracks a response-wide total. `FetchPartHandler` appends every streaming attribute until the FETCH finishes. With `IMAPParserLimits(bodySizeLimit: 64 * 1024 * 1024)` and the default unbounded attribute count, a hostile server can return arbitrarily many 50 MiB BODY sections in one FETCH; every section passes the 64 MiB check while SwiftMail accumulates all of them into one `Data`, so the advertised memory-exhaustion guard remains bypassable.
- **HIGH** `Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift:59` — Parser-limit errors are installed on IDLE connections, but they do not fail closed. During IDLE, an oversized BODY or over-limit attribute count makes `IMAPClientHandler` fire an error; `IdleHandler.errorCaught` only fails its private promise, without finishing the `AsyncStream` or closing the channel, and no task is awaiting that promise until a later DONE/checkpoint. The decoder retains the rejected response and each subsequent network read is appended before the same parse error is raised again (the post-decode buffer cap is skipped on throws), allowing a malicious peer to keep growing memory while the caller's IDLE stream hangs.
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPParserLimits.swift:42` — `literalSizeLimit` is exposed as a configurable maximum, but the pinned `IMAPClientHandler` constructs its byte-to-message processor with `maximumBufferSize: IMAPDefaults.lineLengthLimit` (8,192 bytes), regardless of these parser options or SwiftMail's 1 MiB response buffer limit. For example, a valid 16 KiB mailbox-name literal with `literalSizeLimit: 32 * 1024` succeeds only if delivered complete enough to parse immediately; if network fragmentation leaves more than 8 KiB buffered before the literal completes, it throws `PayloadTooLargeError` even though it is below every configured limit. Thus raised literal limits are not reliably honored.

odrobnik:
@davyd15 Please be sure to comment and resolve the individual review items as you fix them (or tell your agent to do that) For a final merge all threads must be resolved and all CI be green.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
odrobnik, line 31: **HIGH** — `bodySizeLimit` is documented as the maximum body data in one response, but the pinned ResponseParser checks each streaming BODY attribute independently and never tracks the response-wide total. FetchPartHandler appends every streaming attribute until the FETCH finishes. With a 64 MiB limit and the default unbounded attribute count, a hostile server can send arbitrarily many 50 MiB BODY sections in one FETCH; every section passes while SwiftMail accumulates them all into one Data, bypassing the advertised memory guard.

Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
odrobnik, line 61: **HIGH** — Parser-limit errors on an IDLE connection do not fail closed. An oversized BODY or over-limit attribute count makes IMAPClientHandler fire an error, but IdleHandler only fails its private promise; it neither finishes the AsyncStream nor closes the channel, and nothing awaits that promise until a later DONE/checkpoint. The decoder retains the rejected bytes and appends each subsequent network read before throwing again, so a malicious peer can keep growing memory while the caller's IDLE stream hangs.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
odrobnik, line 52: **MEDIUM** — `literalSizeLimit` can be configured above 8 KiB, but the pinned IMAPClientHandler always uses an independent 8,192-byte aggregate buffer cap. A valid 16 KiB mailbox-name literal with `literalSizeLimit: 32 * 1024` can parse if delivered nearly whole, yet fragmented delivery that leaves more than 8 KiB buffered throws PayloadTooLargeError despite remaining below both the configured literal limit and SwiftMail's response buffer limit. Raised literal limits are therefore network-fragmentation dependent and not reliably honored.

Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
odrobnik, line 61: **HIGH** — The pipelined FETCH path also mishandles the newly installed parser-limit errors. PipelinedCommandDispatcher fails all per-request promises, but collectPipelinedFetchResults catches every failure and returns an empty result; the method then removes the dispatcher and reports success without closing the still-active channel. Because the decoder retained the rejected response, later peer writes are appended and rethrow the same error, allowing continued memory growth and leaving subsequent commands on a poisoned connection.

odrobnik (changes requested):
Requesting changes. The TLS-floor propagation and prior inclusive-bound fixes are correct at this head, but the parser hardening still does not meet the PR's hostile-server memory-bounding goal. The three previously reported parser blockers remain, and the pipelined FETCH path also swallows parser-limit failures as an empty success while leaving the poisoned decoder and socket active. Review was performed read-only; builds and tests were not run as requested.

Findings:

- **HIGH** `Sources/SwiftMail/IMAP/IMAPParserLimits.swift:31` — `bodySizeLimit` is documented as the maximum body data in one response, but the pinned ResponseParser checks each streaming BODY attribute independently and never tracks the response-wide total. FetchPartHandler appends every streaming attribute until the FETCH finishes. With a 64 MiB limit and the default unbounded attribute count, a hostile server can send arbitrarily many 50 MiB BODY sections in one FETCH; every section passes while SwiftMail accumulates them all into one Data, bypassing the advertised memory guard.
- **HIGH** `Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift:59` — Parser-limit errors on an IDLE connection do not fail closed. An oversized BODY or over-limit attribute count makes IMAPClientHandler fire an error, but IdleHandler only fails its private promise; it neither finishes the AsyncStream nor closes the channel, and nothing awaits that promise until a later DONE/checkpoint. The decoder retains the rejected bytes and appends each subsequent network read before throwing again, so a malicious peer can keep growing memory while the caller's IDLE stream hangs.
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPParserLimits.swift:42` — `literalSizeLimit` can be configured above 8 KiB, but the pinned IMAPClientHandler always uses an independent 8,192-byte aggregate buffer cap. A valid 16 KiB mailbox-name literal with `literalSizeLimit: 32 * 1024` can parse if delivered nearly whole, yet fragmented delivery that leaves more than 8 KiB buffered throws PayloadTooLargeError despite remaining below both the configured literal limit and SwiftMail's response buffer limit. Raised literal limits are therefore network-fragmentation dependent and not reliably honored.
- **HIGH** `Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift:59` — The pipelined FETCH path also mishandles the newly installed parser-limit errors. PipelinedCommandDispatcher fails all per-request promises, but collectPipelinedFetchResults catches every failure and returns an empty result; the method then removes the dispatcher and reports success without closing the still-active channel. Because the decoder retained the rejected response, later peer writes are appended and rethrow the same error, allowing continued memory growth and leaving subsequent commands on a poisoned connection.

odrobnik:
@davyd15 and please excuse me inundating you with review feedback. This is a new automation that needs some more polishing.

Sources/SwiftMail/IMAP/IMAPConnection.swift
davyd15: Fixed in `ac5ba8b`. `makeConnectionBootstrap` now forwards `minimumTLSVersion` to `makeTLSHandler`.

The root cause was the default value: `= .tlsv1_2` made "forgot to pass the floor" indistinguishable from "chose TLS 1.2", so neither the compiler nor a reviewer could see the difference. `makeTLSHandler` and `IMAPConnection`'s designated initializer no longer default it — omitting it is now a build error. That immediately surfaced a fourth call site (the legacy `useTLS` convenience init) that grep had not found.

Covered by `IMAPTLSFloorEnforcementTests`, which connects a real `IMAPConnection` to a TLS-1.2-pinned endpoint. Verified by reverting the fix: it fails with "an error was expected but none was thrown".

Sources/SwiftMail/IMAP/IMAPServer.swift
davyd15: Fixed in `ac5ba8b`. Both `makeIdleConnection` and `makeNamedConnection` now pass `minimumTLSVersion`.

Same root cause as the implicit-TLS finding: the parameter had a default, so leaving it out looked identical to choosing it. The default is gone, which is what makes this one impossible to reintroduce silently.

Covered by `IMAPSecurityPolicyPropagationTests`, which asserts on the spawned connections themselves rather than on a factory in isolation.

Sources/SwiftMail/IMAP/IMAPServer.swift
davyd15: Fixed in `ac5ba8b`. Both spawned-connection factories now pass `parserLimits`; the parameter no longer has a default either.

Worth noting for the record: this one was the most consequential of the three, since IDLE is exactly where a long-lived connection sits and reads whatever the server chooses to send — with `.default` meaning unbounded.

Covered by `IMAPSecurityPolicyPropagationTests`.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
davyd15: Fixed in `ac5ba8b`. Confirmed against the pinned parser: `guard size < self.bodySizeLimit`, so a body of exactly the documented maximum was rejected.

Rather than redefine "maximum" as "one less than this", the translation to `ResponseParser.Options` now adds one. It saturates at `.max`, where `+ 1` would trap and the value already means unbounded. It lives in `IMAPParserLimits.makeParserOptions(bufferLimit:)` so there is one place to get it right — the previous arrangement, where each call site spelled the options out by hand, is how only the primary connection ended up with limits at all.

Covered by `IMAPParserLimitsTranslationTests`.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
davyd15: Fixed in `ac5ba8b`, same mechanism as the body-size bound. Confirmed against the pinned parser: `guardFetchMiddleAttributeCount()` is called again on `.fetchResponse(.finish)` with the count already incremented, so the effective maximum was `limit - 1` and `messageAttributeLimit: 1` rejected a one-attribute FETCH.

Covered by `IMAPParserLimitsTranslationTests`.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
davyd15: Confirmed and fixed in `32b5536` — this was the sharpest of the findings, thank you.

I could have kept the running total in `FetchPartHandler` and `PipelinedFetchPartHandler`, but then it would hold only for the two consumers that exist today, and only until someone adds a third. `bodySizeLimit` is documented as server-level policy, i.e. a transport promise, so it is now enforced on the transport: `IMAPResponseLimitGuard` sits directly behind the decoder, sums the `byteCount` each `streamingBegin` declares, and resets per message. Acting on the declared count rather than the delivered bytes is deliberate — by the time the bytes arrive the allocation is already decided.

Covered by `IMAPResponseLimitGuardTests`, which drives exactly your scenario (four 30-byte sections against a 100-byte limit). Verified by reverting the accumulation to a per-section check: the test fails.

Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
davyd15: Confirmed and fixed in `32b5536`. You were right on both halves, and the second half is the one I would have missed.

`IMAPResponseLimitGuard` closes the channel on a limit violation — a peer that sends this has proven itself, and the one thing that must not happen is to keep reading from it. But closing alone would have left "fail closed" half true: `IdleHandler` recorded `channelInactive` and `errorCaught` without ever calling `continuation.finish()`, so the socket would be shut and the caller still sitting in `for await`. Both overrides now finish the stream; the error still surfaces through the promise for whoever awaits it.

The close is deliberately narrow — only `ExceededMaximumBodySizeError`, `ExceededMaximumMessageAttributesError` and the new `ExceededResponseBodySizeError`. `errorCaught` sees unrelated failures too, and tearing the connection down for those would change behaviour well outside this feature. `IMAPLimitViolationFailsClosedTests` covers that boundary in both directions.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
davyd15: Confirmed, and this is the one I could not fix — `32b5536` resolves it by removing the promise instead.

`IMAPClientHandler` constructs its `NIOSingleStepByteToMessageProcessor` with `maximumBufferSize: IMAPDefaults.lineLengthLimit`, hardcoded and independent of these options. Honouring a larger literal needs a configurable buffer size there, which the pinned NIOIMAP does not offer. Anything above 8 KiB therefore succeeds or fails depending on how the network fragments it — and a configuration whose outcome depends on packet boundaries is not a limit.

So the initializer now rejects values above `IMAPParserLimits.maximumSupportedLiteralSizeLimit` (= `IMAPDefaults.lineLengthLimit`) rather than accept a bound it cannot keep. Lowering the limit still works and is still useful.

If you would rather see this go the other way — dropping `literalSizeLimit` from the API entirely, or carrying the buffer size upstream into NIOIMAP first — I am happy to change it. It is your call.

Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
davyd15: Confirmed and fixed in `32b5536`. Genuinely new in this round, and a good catch: reported as an empty result, a limit violation looked exactly like "this message has no such section".

`collectPipelinedFetchResults` now rethrows limit violations while still tolerating ordinary per-request failures — swallowing one bad UID is the point of pipelining, but a limit violation is not a property of one request; it means the peer sent more than we allow, and it poisons the shared decoder. It keeps draining the remaining futures before throwing, so NIO's leaked-promise check stays happy, and the cleanup moved into a `defer` so a throw cannot leave the dispatcher installed and `hasActiveHandler` set.

The still-active socket is handled by `IMAPResponseLimitGuard`, which closes on the violation itself.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
davyd15: Same finding as the round-two thread above; answered there. Fixed in `32b5536` — the response-wide total now lives in `IMAPResponseLimitGuard`, behind the decoder.

Sources/SwiftMail/IMAP/IMAPConnection+Connection.swift
davyd15: Same finding as the round-two thread above; answered there. Fixed in `32b5536` — the guard closes the channel, and `IdleHandler` now finishes the `AsyncStream` on `channelInactive` and `errorCaught`.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
davyd15: Same finding as the round-two thread above; answered there. Resolved in `32b5536` by rejecting values above 8 KiB instead of promising a bound the pinned `IMAPClientHandler` cannot keep.

davyd15:
Nothing to excuse — the automation earned its keep. Nine distinct findings across three rounds, and every one of them was real: I checked each against the code before touching it, and none turned out to be a false positive. Two were the kind that tests do not catch. The TLS floor was silently inert on port 993, the only port that matters for IMAPS — while a test named *"Raising the floor to TLS 1.3 makes a downgrade impossible"* sat there green, because it asserted that a factory returns what it is handed, and the defect was that nobody handed it the value. If it is useful signal while you tune the thing: the findings got sharper each round rather than repeating themselves, and the pipelined-FETCH one was genuinely new.

The noise on this branch was mine, though: I pushed twice without answering the threads, the second time for a lint fix only. That is what turned one review into three. Threads get answered before anything lands here from now on.

**All findings are addressed at `32b5536`.** Every fix has a test that fails without it — I reverted each one to check, including the TLS floor and the response-wide body total. Ten threads answered and resolved above; the two `literalSizeLimit` ones I left open on purpose, see below.

Two things worth your attention rather than mine:

**The fixes reach past the limits type.** `bodySizeLimit` needed response-wide accounting that the pinned parser does not do and `FetchPartHandler` defeats. I could have counted the total in the two handlers that accumulate, but it would hold only for those two and only until a third appears — so it is enforced on the transport instead (`IMAPResponseLimitGuard`, directly behind the decoder). Failing closed then reached `IdleHandler` (which recorded `channelInactive` and `errorCaught` without ever finishing the `AsyncStream`) and the pipelined FETCH path (which reported limit violations as an empty success). That is a broader change than "make the existing knobs configurable". **If you would rather have the parser limits as their own PR so the TLS floor can land on its own, say the word and I will split it** — I would not restructure your PR queue unasked.

**`literalSizeLimit` is the one I could not honour.** `IMAPClientHandler` hardcodes `maximumBufferSize: IMAPDefaults.lineLengthLimit`, so above 8 KiB the outcome depends on how the network fragments the literal. Rather than document a bound that packet boundaries decide, the initializer now rejects those values. Dropping the property entirely, or carrying a configurable buffer size upstream into NIOIMAP first, are both defensible alternatives — your call, and I left those two threads open for it.

Verified locally: 346 tests pass, `swiftlint --strict` clean across 264 files.

Sources/SwiftMail/IMAP/IMAPParserLimits.swift
odrobnik: should be good for now.

odrobnik: should be good for now.

davyd15:
One more, and this one was hiding in plain sight: **the Swift workflow has been failing on this branch since `ac5ba8b`** — my first round of fixes. I only checked once the threads were closed.

It read as *cancelled* rather than failed, which is why neither of us caught it: `build-macos` hits its `timeout-minutes: 10` at "Test (macOS)" on every head. `main` and other branches are green, so it is ours.

Cause: `try? group.syncShutdownGracefully()` called straight from a swift-testing test blocks a cooperative-pool thread; on a core-constrained runner that violates the pool's forward-progress guarantee and deadlocks the run. Locally it never showed — 346 tests in 1.6 seconds on a machine with cores to spare.

Which is the annoying part: #179 already hit this, documented it in `IMAPResponseBufferLimitTests` ("a 7-minute hang to the job timeout was observed"), and fixed it by dispatching the blocking call to a non-cooperative GCD queue. I wrote the same trap directly next to the note warning about it.

`66c7433` uses that pattern, extracted to a shared `shutDownGracefully(_:)` rather than copied a third time, and the new suites adopt `.serialized` + `.timeLimit(.minutes(1))` like the existing ones — a hanging test should fail in a minute and say so, rather than eat the job timeout and report as "cancelled".

Verified locally: 346 tests, `swiftlint --strict` clean. Watching CI on this head; I will confirm here once macOS and iOS are green rather than leave you to check.

davyd15:
I need to ask for help on CI, before I burn more of your review runs guessing.

**`build-macos` has never been green on this branch** — it hits `timeout-minutes: 10` at "Test (macOS)" on every head since `ac5ba8b`. It reports as *cancelled* rather than failed, which is why it went unnoticed for so long. `main` does the whole job in ~5 minutes, so this is ours.

What the logs show: it is **not** one hanging test. Roughly 27 of 347 complete, then the run stalls for ~7 minutes until the timeout — and the tests left unfinished include ones with no relation to this PR (`"ASCII display name behaves like description"`). That reads like cooperative-pool starvation rather than a deadlock in any single place.

**It does not reproduce locally.** 347 tests in 1.6s, and still 1.6s with `LIBDISPATCH_COOPERATIVE_POOL_STRICT=1`. This machine has cores to spare, which I suspect is exactly the difference.

Four attempts, none of which helped:

1. Removed `syncShutdownGracefully()` from the new tests, using the GCD-dispatch pattern #179 established.
2. Removed `bind(...).wait()` from the test server — nothing else in the suite calls `.wait()`; I had added the only three.
3. Replaced the real-socket TLS server entirely with an in-memory `EmbeddedChannel` handshake — no threads, no sockets, no event loop group. Still stalls.
4. Added `.serialized` + `.timeLimit(.minutes(1))` to the two new suites that construct servers, matching every pre-existing suite that does.

**One observation I will offer as a lead rather than a diagnosis**, since I could not confirm it: `SMTPServer.deinit` calls the blocking `syncShutdownGracefully()`, while `IMAPServer.deinit` dispatches to a `Task`. This PR adds six server constructions across two test suites. That is the only mechanism I can find by which new *tests* would starve the pool for *unrelated* ones — but attempt 4 should then have fixed it, and it did not, so I am probably still wrong somewhere.

Have you seen this before on this runner? And is there a variant you would prefer: dropping the new suites and keeping only what does not construct servers, or something else entirely? I would rather ask than push a fifth guess at it.

Unrelated, for completeness: `build-ios` failed on the last two runs with `iOS 26.0 is not installed` from `xcodebuild` — runner image, not this branch.

Everything else stands: all findings addressed, all twelve threads resolved, SwiftLint green, Linux and Android green, 347 tests and `swiftlint --strict` clean locally.

odrobnik:
Let's See what Fable Ultra finds... stand by...

odrobnik:
Thanks for this contribution — both changes are solid and they're being merged. The merge happens via #186, which contains this PR's commits verbatim (your authorship intact; GitHub will mark this PR merged automatically when #186 lands) plus follow-up commits that got CI green and tightened a few things found in review:

- **The CI hangs you fought were real, and only one-third yours.** Your diagnosis in `66c7433` was on the right track but incomplete: the blocking `syncShutdownGracefully()` lives in `SMTPServer.deinit`, so `.serialized` on the new suites couldn't fix it — three *suites* concurrently releasing servers (your new `MailTLSConfigurationTests` plus the two pre-existing SMTP suites) park all three cooperative-pool threads on a 3-core runner, and the shutdown's own completion callback needs one of those threads. Thread samples from a hung runner confirmed it. Fixed in the library: both server deinits now use the non-blocking callback form.
- **Linux hung separately**: `EmbeddedChannel.finish()` after a completed TLS handshake does `close().wait()`, and NIOSSL's graceful shutdown waits for a close_notify the stopped byte pump can never deliver (macOS completes the close inline, which is why you never saw it). The teardown now pumps the close_notify exchange and advances embedded time.
- **build-ios was pure infrastructure**: the macos-latest image rotated on July 15 and stranded the pinned Xcode 26.0 without its iOS platform — unmodified main failed identically. The workflow now tracks latest-stable.
- **One functional fix worth knowing about**: `ResponseDecoder` wraps every parser error in `IMAPDecoderError` before it reaches the pipeline, so `isLimitViolation` never matched a production-raised violation — the fail-closed close only ever fired for the bare error types the tests injected. It now unwraps the wrapper (and also matches `ExceededLiteralSizeLimitError`), with an end-to-end test that drives raw wire bytes through the real decoder.

The TLS floor and parser-limit design, the propagation to IDLE/named connections, and the no-defaults principle all survive unchanged — good work, and the write-ups in your commit messages made the archaeology much easier.

PR #193, opened 2026-07-26
Replay RFC 2971 ID after every authentication when configured

heymi:
## Motivation

Some IMAP servers — notably NetEase (163.com / 126.com / yeah.net), one of the largest mail providers in China — reject every `SELECT` on an authenticated connection that has not identified itself via the RFC 2971 `ID` command, answering `NO Unsafe Login. Please contact kefu@188.com for help`.

Callers can already send `ID` manually on the primary connection via `IMAPServer.id(_:)`. But connections the server opens internally authenticate **without** `ID` and are permanently rejected by such servers:

- the dedicated IDLE connection (`idle(on:)` → `makeIdleConnection` → authenticate → `SELECT`), so mailbox IDLE never establishes;
- IDLE recovery reconnects;
- transparent re-authentication in `executeCommand` after the primary connection is dropped, so a recycled connection stalls until the caller reconnects explicitly.

## Change

- `IMAPServer.setClientIdentification(_:)` (new public API) stores an optional `Identification`.
- The internal `Authentication` enum becomes a struct carrying the identification alongside the auth method; `authenticate(on:)` replays `ID` right after every successful authentication on whichever connection it authenticated — all internal call sites are covered without modification.
- The three public auth methods replay on the primary connection as well, so one configuration call covers every connection for the server's lifetime.
- Guards: the replay is skipped when the connection does not advertise the `ID` capability (checked against the post-auth capability snapshot), and it is best-effort — a failed `ID` never fails the authentication itself. **Behavior is unchanged unless the new API is used.**

## Verification

- `swift build` clean on this branch.
- Verified downstream (Orvia mail client, which vendors this patch on 1.5.2 and 1.6.4) against a real NetEase 163 account: with the identification configured, mailbox IDLE establishes and pushes arrive in seconds; without it, the dedicated IDLE connection's `SELECT` is rejected as described.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPServer.swift
+++ b/Sources/SwiftMail/IMAP/IMAPServer.swift
@@ -71,6 +71,12 @@ public actor IMAPServer {
     /// Authentication configuration for spawning new connections.
     var authentication: Authentication?
 
+    /// RFC 2971 client identity replayed after every successful
+    /// authentication on every connection this server opens (primary,
+    /// dedicated IDLE connections, and transparent re-authentication).
+    /// Set it before calling `login`/`authenticatePlain`/`authenticateXOAUTH2`.
+    var clientIdentification: Identification?
+
     /** The list of all mailboxes with their attributes */
     public private(set) var mailboxes: [Mailbox.Info] = []
 
@@ -113,13 +119,18 @@ public actor IMAPServer {
         let handle: IMAPNamedConnection
     }
 
-    enum Authentication {
-        case login(username: String, password: String)
-        case plain(username: String, password: String)
-        case xoauth2(email: String, accessTokenProvider: @Sendable () async throws -> String)
+    struct Authentication {
+        enum Method {
+            case login(username: String, password: String)
+            case plain(username: String, password: String)
+            case xoauth2(email: String, accessTokenProvider: @Sendable () async throws -> String)
+        }
+
+        let method: Method
+        var identification: Identification?
 
         func authenticate(on connection: IMAPConnection) async throws {
-            switch self {
+            switch method {
                 case .login(let username, let password):
                     try await connection.login(username: username, password: password)
                 case .plain(let username, let password):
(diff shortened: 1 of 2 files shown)


Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -128,6 +139,14 @@ public actor IMAPServer {
                     let accessToken = try await accessTokenProvider()
                     try await connection.authenticateXOAUTH2(email: email, accessToken: accessToken)
             }
+            // RFC 2971: some servers (e.g. NetEase 163/126) reject SELECT on any
+            // authenticated connection that has not identified itself, so the
+            // stored identity is replayed after every authentication. Servers
+            // that do not advertise the ID capability are skipped, and a
+            // failed ID must never fail the authentication itself.
+            if let identification, connection.capabilitiesSnapshot.contains(.id) {
```
odrobnik, line 147: **MEDIUM** — A server advertises `AUTH=XOAUTH2` before authentication, advertises `ID` only after authentication, and returns a normal successful AUTHENTICATE response without a CAPABILITY response code. `authenticateXOAUTH2` then retains the pre-authentication capability snapshot, so this guard sees no `.id` and skips the replay on the primary connection and on every IDLE/named/recovery connection. A subsequent SELECT on a server enforcing identification is rejected with `NO Unsafe Login` even though the caller configured `setClientIdentification`.

Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -128,6 +139,14 @@ public actor IMAPServer {
...
+                _ = try? await connection.id(identification)
```
odrobnik, line 148: **MEDIUM** — If the server advertises `ID`, accepts authentication, and then closes the socket during `ID` (or the command times out), `IMAPConnection.executeCommand` recycles the transport and clears `isSessionAuthenticated`; this `try?` nevertheless makes `Authentication.authenticate` return success. `connection(named:)` can therefore return a disconnected, unauthenticated handle, while an IDLE setup/recovery proceeds to SELECT after an unauthenticated auto-reconnect and fails. The identical blanket suppression in `identifyPrimaryConnectionIfNeeded` also lets the public login/authenticate methods return success with the primary connection already disconnected.

odrobnik (changes requested):
Requesting changes for two correctness issues in the replay path. XOAUTH2 can test a stale pre-authentication capability set and skip `ID`, and blanket suppression of `ID` errors can report authentication success after the command executor has already disconnected and cleared the session. Together these violate the promised coverage of every authentication path and create unusable “authenticated” connections. The review was read-only; builds and tests were not run as requested, and this PR adds no automated coverage for the new replay behavior.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer.swift:147`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer.swift:148`

odrobnik:
@heymi the ball is back in your court, please address the lint issue and the review findings.

PR #198, opened 2026-07-28
Harden pipelined part fetch: content-verified routing, unsolicited FETCH suppression, failure propagation

jdudley:
## Problem

Integrating `fetchPartsPipelined` into a mail client surfaced two defects in the pipelined response path, both of which corrupt data silently.

**1. Unsolicited FETCH responses shift part routing.** `PipelinedCommandDispatcher` routes every untagged FETCH response to the oldest unfinished command. RFC 3501 §7.4.2 allows the server to broadcast unsolicited FETCH responses at any time — the common case being `* n FETCH (FLAGS (\Seen))` when another connection marks a message read mid-burst. That attribute-only group consumed a command's routing slot: every subsequent part's bytes shifted one command over (the HTML part received the plain-text part's bytes, attachments received the HTML), the last part came back empty, and all of it was returned as a successful fetch.

**2. Failures are swallowed; a poisoned channel is reused.** `collectPipelinedFetchResults` caught every per-command failure and returned partial (or empty) results. A batch where *every* command failed returned `[]` — indistinguishable from "this message has no such sections" — even though the API doc promised a throw. Worse, the whole-batch timeout only failed the promises: unlike serial execution (which classifies timeout as recyclable and disconnects), the connection stayed installed with literals for the timed-out tags still streaming, so the next command could read another command's stale data as its own response.

A smaller parity gap: the batch shared a single 60s timeout, while the serial path it replaced allowed 60s *per part* — a large multipart message on a slow link timed out pipelined where serial fetching succeeded.

## Changes

- **Group-based routing with proof-of-body**: untagged FETCH responses are processed as groups (`.start` … `.finish`), held back (start + attributes only — bounded and small) until the group proves it answers a pipelined command by streaming body data. Attribute-only groups are discarded at `.finish`.
- **Content-verified binding**: commands register their expected UID and section (mirroring `FetchMessagePartCommand`'s encoding); a group binds to the entry matching its UID attribute and streamed section when the server provides them, falling back to send order otherwise (existing behavior — the four pre-existing routing tests pass unchanged). A group carrying several sections routes each section to its own command; a body-bearing group for an unrequested UID is dropped; a UID mismatch discovered after binding logs a warning.
- **Failure propagation**: the drain step now records the first failure separately from parser-limit violations. All-commands-failed throws that failure (making the documented behavior true). Failures classified recyclable by the existing `shouldRecycleConnection` (timeout, channel drop) disconnect the connection before results are returned — including the partial-results case — mirroring serial command execution. Send-path failures recycle the same way.
- **Scaled batch timeout**: `fetchPartsPipelined` gains `timeoutSeconds: Int? = nil`, defaulting to 60s + 30s per additional part, capped at 300s.

## Tests

- New `PipelinedFetchHardeningTests`: the unsolicited flag-update regression, out-of-order section routing by UID+section, a single group satisfying multiple commands, unrequested-UID body-group dropping, all-failed/partial/limit-violation drain policy, and the timeout formula.
- All four pre-existing dispatcher routing tests pass unmodified, pinning order-fallback compatibility.
- Full suite: 366 tests in 51 suites pass; `swiftlint lint --strict` clean.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
@@ -4,11 +4,19 @@
 // IMAP RFC 3501 §5.5 allows clients to send multiple commands without waiting for responses.
 // The server processes commands in order and sends tagged responses (A001 OK, A002 OK, etc.)
 // so responses can be matched to commands by tag. Untagged responses (e.g., * FETCH data)
-// carry no tag; they arrive in command order and are routed to the oldest command whose
-// untagged response has not yet finished, advancing on each response's `.finish`. A server
-// may stream data for several pipelined commands before sending any tagged OK (RFC 3501
-// §5.5), so advancing only on the tagged OK would misdeliver a later command's data to an
-// already-finished earlier handler and silently drop it.
+// carry no tag; they arrive grouped per message (`.start` … `.finish`) and are routed by
+// content when the group identifies itself (UID attribute, streamed body section) and by
+// send order otherwise. A server may stream data for several pipelined commands before
+// sending any tagged OK (RFC 3501 §5.5), so advancing only on the tagged OK would misdeliver
+// a later command's data to an already-finished earlier handler and silently drop it.
+//
+// Unsolicited FETCH responses are a hard requirement here: RFC 3501 §7.4.2 lets the server
+// broadcast flag changes (`* n FETCH (FLAGS (\Seen))`) at any time, including interleaved
+// into a pipelined burst. Such a group carries attributes only — no streamed body data — and
+// must not consume a pending command's routing slot: doing so shifts every subsequent part's
+// bytes one command over and the corruption is cached as a successful fetch. Groups are
+// therefore held back until they prove they belong to a command by streaming body data;
+// attribute-only groups are discarded at their `.finish`.
 //
 // This handler sits in the NIO pipeline during a pipelined batch. It maintains an ordered
 // registry of (tag → PipelinedHandler) and routes responses accordingly.
@@ -27,24 +35,55 @@ final class PipelinedCommandDispatcher: ChannelInboundHandler, RemovableChannelH
     private let lock = NIOLock()
 
     /// A registered pipelined command awaiting its responses. `finishedUntagged` flips
-    /// once this command's untagged FETCH response has fully arrived (its `.finish`), so
-    /// subsequent untagged data routes to the next command even before any tagged OK.
+    /// once this command's untagged FETCH data has fully arrived, so subsequent untagged
+    /// data routes to the next command even before any tagged OK. `uid` and
+    /// `expectedSection` enable content-verified routing when the server's response
+    /// identifies itself; entries registered without them fall back to send order.
     private struct PendingCommand {
         let tag: String
         let handler: any PipelinedHandler
+        let uid: UID?
+        let expectedSection: SectionSpecifier?
         var finishedUntagged: Bool
     }
 
-    /// Ordered registry — insertion order matches send order. Untagged FETCH responses
-    /// route to the oldest entry whose untagged response has not finished.
+    /// The untagged FETCH group currently arriving (one `.start` … `.finish` span).
+    private enum UntaggedGroup {
+        case idle
+        /// `.start` seen but no body data yet. The group's responses are held back
+        /// (start plus simple attributes — bounded and small) until it proves it
+        /// belongs to a pipelined command by streaming body data.
+        case pending(held: [FetchResponse], uid: UID?)
+        /// Group bound to a command; body chunks stream straight through. A server
+        /// may satisfy several pipelined commands for one message inside a single
+        /// group, so `touchedTags` tracks every entry the group delivered to.
+        case bound(currentTag: String, touchedTags: Set<String>, uid: UID?)
+        /// Group streams body data no pending command asked for; swallowed to `.finish`.
+        case dropping
+    }
+
+    /// Ordered registry — insertion order matches send order.
     private var entries: [PendingCommand] = []
+    private var group: UntaggedGroup = .idle
 
     private let logger = Logger(label: "com.cocoanetics.SwiftMail.PipelinedDispatcher")
 
     /// Register a handler for a command tag. Must be called in send order.
-    func register(tag: String, handler: any PipelinedHandler) {
+    /// Pass `uid` and `expectedSection` to enable content-verified routing.
+    func register(
+        tag: String,
+        handler: any PipelinedHandler,
+        uid: UID? = nil,
+        expectedSection: SectionSpecifier? = nil
+    ) {
         lock.withLock {
-            entries.append(PendingCommand(tag: tag, handler: handler, finishedUntagged: false))
+            entries.append(PendingCommand(
+                tag: tag,
+                handler: handler,
+                uid: uid,
+                expectedSection: expectedSection,
+                finishedUntagged: false
+            ))
         }
     }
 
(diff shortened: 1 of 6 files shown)


Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
```diff
@@ -84,27 +123,208 @@ final class PipelinedCommandDispatcher: ChannelInboundHandler, RemovableChannelH
...
+        guard !candidates.isEmpty else { return nil }
+
+        if let uid {
+            let uidMatches = candidates.filter { entries[$0].uid == uid }
+            if !uidMatches.isEmpty {
+                if let section,
+                   let exact = uidMatches.first(where: { entries[$0].expectedSection == section }) {
+                    return exact
+                }
+                return uidMatches.first
```
odrobnik: **HIGH** — When both UID and section are present but no pending command matches the pair, `bindableEntryIndex` falls back to the first command with the same UID instead of rejecting the group. For requests `(UID 9, BODY[1])` and `(UID 9, BODY[2])`, an unsolicited `* n FETCH (UID 9 BODY[99] {4}\r\nEVIL)` is delivered to the BODY[1] handler and marks it finished; the later legitimate BODY[1] response can then be shifted to BODY[2], and both tagged OKs return corrupted data as successes. A known `(uid, section)` mismatch must be dropped or fail closed rather than degrade to a UID-only match.

Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
```diff
@@ -84,27 +123,208 @@ final class PipelinedCommandDispatcher: ChannelInboundHandler, RemovableChannelH
...
+        switch group {
+            case .pending(var held, var uid):
+                if case .uid(let nioUID) = attribute {
+                    uid = UID(nio: nioUID)
+                }
+                held.append(response)
+                group = .pending(held: held, uid: uid)
+            case .bound(let currentTag, let touchedTags, _):
+                if case .uid(let nioUID) = attribute {
+                    verifyBoundUID(UID(nio: nioUID), currentTag: currentTag)
```
odrobnik: **HIGH** — The parser permits attributes in either order, including `BODY[...] {n}` followed by `UID`, but the dispatcher commits literal bytes before a later UID is known and only logs if that UID disproves the binding. With pending `(UID 9, BODY[1])` and `(UID 10, BODY[1])`, if the UID 10 response arrives first as `BODY[1] ... UID 10`, its bytes bind by section/send order to the UID 9 handler; a subsequent UID 9 response can bind to UID 10, so tagged OKs successfully return swapped bodies. Once a post-literal UID mismatch is detected, the batch must not be allowed to succeed with the already-misrouted bytes.

odrobnik (changes requested):
Requesting changes: the content router still has two fail-open paths that can silently assign body bytes to the wrong request, and the advertised scaled timeout is not applied by the primary `IMAPServer` public API. The new tests cover mismatched UIDs only when the UID precedes the literal; they do not exercise an unrequested section for a requested UID or a UID attribute that follows the literal.

- **HIGH** `Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift:308`
- **HIGH** `Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift:178`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer+Fetch.swift` — The timeout-scaling requirement is wired only into `IMAPNamedConnection.fetchPartsPipelined`; the equally public and primary `IMAPServer.fetchPartsPipelined` still calls `executePipelinedFetchParts(requests:)`, whose default remains 60 seconds. A client using `IMAPServer` to fetch a 10-part message therefore still times out the entire batch at 60 seconds, even though this PR advertises a scaled/capped default specifically to preserve serial-fetch parity. Apply the same timeout parameter/default policy at this public entry point.

odrobnik:
@jdudley please address the issues found in review

Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
jdudley: Fixed in dc29553. UID-aware routing now requires an exact section match; a requested UID paired with an unrequested section creates a typed routing-integrity failure, fails every pending handler, and forces connection recycle. Added the unrequestedSectionForRequestedUIDFailsBatch regression.

Sources/SwiftMail/IMAP/IMAP/Handler/PipelinedCommandDispatcher.swift
jdudley: Fixed in dc29553. A late UID is now validated against every tag that already received body bytes; any mismatch fails all pending handlers. The drain treats this typed violation as batch-fatal even after another request completed, then recycles the connection. Added post-literal and partial-success regressions.

jdudley:
Addressed all three review findings in dc29553. In addition to the two fail-closed routing fixes covered in the resolved threads, the scaled timeout default is now centralized in IMAPConnection.executePipelinedFetchParts, and both IMAPServer and IMAPNamedConnection expose the same optional timeout override. Validation: swift test --filter PipelinedFetch (23 tests passed) and swiftlint lint --strict (0 violations).

odrobnik:
MissionControl review timed out for head `dc29553cb1e9`. Auto-merge remains blocked; human review is required.

odrobnik:
MissionControl review timed out for head `dc29553cb1e9`. Auto-merge remains blocked; human review is required.

odrobnik (approved):
All three findings from the prior review are resolved correctly.

**HIGH (line 308) — unrequested section for a requested UID:** `sectionDecision` now returns `.routingViolation` when no pending command's `expectedSection` matches the server's section, and `failAllPending` is called. The `unrequestedSectionForRequestedUIDFailsBatch` regression test covers this path.

**HIGH (line 178) — post-literal UID mismatch:** `validateBoundUID` fires when a `.simpleAttribute(.uid)` arrives inside an already-bound group. It compares the reported UID against every touched entry's registered UID and calls `failAllPending` on any mismatch. The `postLiteralUIDMismatchFailsBatch` regression test covers this path.

**MEDIUM (IMAPServer+Fetch.swift) — scaled timeout not wired at the public entry point:** Both `IMAPServer.fetchPartsPipelined` and `IMAPNamedConnection.fetchPartsPipelined` now accept `timeoutSeconds: Int? = nil` and pass it through to `executePipelinedFetchParts`, which resolves the default via `IMAPConnection.defaultPipelinedFetchTimeout`. The `defaultTimeoutScalesWithPartCount` test pins the formula.

Additional checks with no findings:

- Section-specifier equality is correct: production registration and `FetchMessagePartCommand.toTaggedCommand` both construct `SectionSpecifier(part: SectionSpecifier.Part(section.components))`, matching the NIOIMAP parser's output.
- `shouldRecycleConnection` returns false for `PipelinedFetchRoutingError`; the routing-violation arm in `resolvePipelinedFetchDrain` calls `disconnectBody` before throwing, so the channel is recycled on that path.
- `markTouchedFinished` delivers `.finish` only to `currentTag` and marks all `touchedTags` as `finishedUntagged`. Non-current touched handlers do not receive `.finish` (`didFinishPart` stays false), but `processTaggedResponse` resolves the promise from `partData` without consulting `didFinishPart`, and `finishedUntagged` prevents any further routing to those handlers — so no stale-data or unresolved-promise risk.
- `failAllPending` sets `group = .idle`; the caller in the routing-violation path then sets `group = .dropping`. The override is harmless: both states discard remaining bytes, and the subsequent `.finish` resets to `.idle` either way.
- `UID FETCH` guarantees the server includes UID as a `simpleAttribute`, so content-verified routing works as designed in production.

👍 No findings.

PR #202, opened 2026-07-30
Fix byte offsets when splitting EML headers

### LINKED ISSUE #201: EMLParser uses character offsets to slice raw message bytes
## Summary

`EMLParser.splitHeadersAndBody(from:rawData:)` locates the header/body separator in a `String`, then passes `String.distance(...)` to `Data.dropFirst`. A Swift `Character` count is not a byte offset: CRLF can be represented as one grapheme while occupying two bytes, and non-ASCII headers can occupy multiple UTF-8 bytes.

## Impact

The MIME body can begin at the wrong byte. Base64 attachment decoding can then fail or return corrupted data, so otherwise valid attachments appear invalid to consumers.

## Reproduction

Parse a CRLF-separated multipart message containing a Base64 attachment and assert:

```swift
message.attachments[0].decodedData() == Data("Hello World".utf8)
```

Before the proposed fix the decoded result contains an extra byte; after the fix it matches exactly.

## Proposed fix

Search for the blank-line separator directly in `rawData`, decode only the header byte range into `String`, and slice the body at `range.upperBound`.

## Validation

- Added a regression assertion for exact Base64 attachment contents
- Full test suite passes: 404 tests in 52 suites

kolisko:
## What changed

- find the EML header/body separator in the original `Data`
- decode only the header byte range
- add a regression assertion for exact Base64 attachment decoding

## Root cause

The parser used a Swift `String` character distance as a raw `Data` byte offset. CRLF and multi-byte UTF-8 content make those units diverge, so the body could start too early and corrupt MIME transfer decoding.

Fixes #201

## Validation

- `swift test`
- 404 tests in 52 suites passed

### DIFF
--- a/Sources/SwiftMail/MIME/EMLParser+Multipart.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+Multipart.swift
@@ -160,7 +160,9 @@ extension EMLParser {
         let childPath = sectionPath.isEmpty ? [partNumber] : sectionPath + [partNumber]
 
         let partData = Data(rawPart.utf8)
-        let (partHeaders, partBody) = splitHeadersAndBody(from: rawPart, rawData: partData)
+        guard let (partHeaders, partBody) = try? splitHeadersAndBody(rawData: partData) else {
+            return []
+        }
         let headers = parseHeaders(partHeaders)
 
         let partContentType = headers["content-type"] ?? "text/plain"
--- a/Sources/SwiftMail/MIME/EMLParser+Multipart.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+Multipart.swift
@@ -49,118 +49,154 @@ extension EMLParser {
             )]
         }
 
-        let utf8Body = String(data: bodyData, encoding: .utf8)
-        let asciiBody = String(data: bodyData, encoding: .ascii)
-        guard let bodyString = utf8Body ?? asciiBody else {
-            return []
-        }
-
-        let rawParts = splitMultipartByBoundary(bodyString: bodyString, boundary: boundary)
+        let rawParts = splitMultipartByBoundary(bodyData: bodyData, boundary: boundary)
         return rawParts.enumerated().flatMap { index, rawPart -> [MessagePart] in
             buildMultipartChild(rawPart: rawPart, index: index, sectionPath: sectionPath)
         }
     }
 
-    /// Split a multipart body string into raw part strings using the boundary
-    /// delimiter. Returns an array of raw (header+body) part strings.
-    static func splitMultipartByBoundary(bodyString: String, boundary: String) -> [String] {
-        let delimiter = "--\(boundary)"
-        var rawParts: [String] = []
-        var searchStart = bodyString.startIndex
+    /// Split a multipart body into opaque byte slices using valid boundary lines.
+    /// The preamble, epilogue, and CRLF framing around each part are discarded.
+    static func splitMultipartByBoundary(bodyData: Data, boundary: String) -> [Data] {
+        let marker = Data("--\(boundary)".utf8)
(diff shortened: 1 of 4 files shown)


odrobnik (changes requested):
The byte-offset change fixes the CRLF/base64 regression for UTF-8/ASCII messages, but it does not complete issue #201’s requirement to decode only the header byte range.

`EMLParser.parse` still converts the entire raw message to UTF-8/ASCII before calling the new byte-based splitter. Consequently, a standards-compliant 8bit or binary body containing non-UTF-8 bytes is rejected before the changed code can preserve it. Please locate the top-level separator from `Data` first and decode only the header bytes (and keep body bytes opaque); add coverage with a non-UTF-8 8bit/binary body.

Findings:

- **HIGH** `Sources/SwiftMail/MIME/EMLParser.swift` — Parse an EML with ASCII headers such as `Content-Type: text/plain; charset=iso-8859-1` and `Content-Transfer-Encoding: 8bit`, followed by CRLFCRLF and a valid ISO-8859-1 body containing byte 0xE9 (or a binary body containing 0xFF). `EMLParser.parse` first attempts to decode the complete `Data` as UTF-8/ASCII, both conversions fail, and it throws `EMLParserError.invalidData`; `splitHeadersAndBody` is never reached, so the raw body cannot be returned as a `MessagePart` despite issue #201 requiring only header bytes to be decoded.

kolisko:
Addressed in `a4399c7`.

`EMLParser.parse` now locates the top-level header/body separator in the original `Data`, decodes only the header bytes, and preserves the body as opaque bytes. I added regression coverage for an ISO-8859-1 `8bit` body containing `0xE9` and an opaque binary body containing `0xFF`.

The complete suite passes locally: 406 tests in 52 suites.

Sources/SwiftMail/MIME/EMLParser+Multipart.swift
```diff
@@ -160,7 +160,9 @@ extension EMLParser {
         let childPath = sectionPath.isEmpty ? [partNumber] : sectionPath + [partNumber]
 
         let partData = Data(rawPart.utf8)
-        let (partHeaders, partBody) = splitHeadersAndBody(from: rawPart, rawData: partData)
+        guard let (partHeaders, partBody) = try? splitHeadersAndBody(rawData: partData) else {
```
odrobnik: **HIGH** — Parse a message with ASCII top-level headers declaring `multipart/mixed; boundary="b"`, then include a child with ASCII MIME headers, `Content-Type: text/plain; charset=iso-8859-1`, `Content-Transfer-Encoding: 8bit`, and body bytes `63 61 66 E9` before the closing boundary (the same occurs with a binary child containing `FF`). `parseMultipart` first tries to decode the entire outer body as UTF-8/ASCII; both conversions fail and it returns `[]`, so the changed `splitHeadersAndBody(rawData:)` call is never reached. `Message(emlData:)` succeeds with no parts, silently dropping every body/attachment instead of preserving the original child bytes required by issue #201. Split multipart boundaries over `Data` and pass original byte slices into the child splitter; cover both multipart 8bit and binary bodies.

odrobnik (changes requested):
Requesting changes: the top-level header split now fixes the CRLF/base64 byte-offset regression and preserves opaque bytes for single-part messages, but multipart parsing still converts the entire MIME body to UTF-8/ASCII before the new byte splitter can run. A valid multipart containing a non-UTF-8 8bit or binary child is therefore accepted with zero parts, silently losing its body and attachments. Multipart boundary and child slicing must operate on the original `Data`, with multipart non-UTF-8 coverage.

- **HIGH** `Sources/SwiftMail/MIME/EMLParser+Multipart.swift:163`

PR #206, opened 2026-08-05
Populate message info from selective header fields

jdudley:
## Summary

- populate missing standard MessageInfo fields from BODY[HEADER.FIELDS (...)] literals
- preserve parsed ENVELOPE values when both representations are present, including explicit empty strings
- add regression coverage for no-ENVELOPE fallback, ENVELOPE authority, and legacy 8-bit header bytes

## Motivation

FetchMessageInfoOptions permits callers to omit ENVELOPE and request selected headers, but the handler currently filters standard headers as presumed ENVELOPE duplicates. That leaves subject, sender, recipients, date, and message IDs empty when a caller intentionally uses header fields as a protocol-safe fallback.

## Validation

- swift test — 435 tests passed
- swiftlint lint --strict --no-cache — 0 violations

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift
@@ -121,6 +121,8 @@ final class FetchMessageInfoHandler: BaseIMAPCommandHandler<[MessageInfo]>, IMAP
             guard let index = currentMessageIndex() else { return }
             var header = self.messageInfos[index]
 
+            Self.applyMissingStandardHeaders(allHeaders, to: &header)
+
             if let references = referencesValue, !references.isEmpty {
                 let parsed = Self.parseMessageIDs(from: references)
                 header.references = parsed.isEmpty ? nil : parsed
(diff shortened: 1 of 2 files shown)


Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift
```diff
@@ -131,6 +133,28 @@ final class FetchMessageInfoHandler: BaseIMAPCommandHandler<[MessageInfo]>, IMAP
...
 
+    /// Populate standard message fields from a requested header literal when
+    /// ENVELOPE was omitted or left a field empty. ENVELOPE values stay
+    /// authoritative when both representations are present.
+    private static func applyMissingStandardHeaders(
+        _ fields: [String: String],
+        to header: inout MessageInfo
+    ) {
+        let parsed = EMLParser.buildMessageInfo(from: fields)
+        if header.subject?.isEmpty != false { header.subject = parsed.subject }
```
odrobnik: **LOW** — If an IMAP server returns the ENVELOPE subject as an empty quoted string "" (valid per RFC 3501, and possible when a message carries an encoded-word subject that decodes to the empty string via decodeMIMEHeader), applyEnvelope sets header.subject = "". The condition `header.subject?.isEmpty != false` then evaluates to true and the ENVELOPE value is overwritten with the header-literal subject, violating the stated invariant that ENVELOPE is authoritative. Using `header.subject == nil` instead closes the gap.

odrobnik (changes requested):
The overall approach is sound: `applyMissingStandardHeaders` is called inside the existing lock, ENVELOPE-first ordering is guaranteed by the IMAP response structure (simpleAttribute fires before streamingEnd), references handling is unchanged, and the two new tests cover the primary fallback and non-override paths.

One low-severity correctness edge case: the guards for `subject` and `from` use `?.isEmpty != false` instead of `== nil`. This means any ENVELOPE-sourced empty string ("" rather than NIL) gets silently overwritten by the header-literal value, contradicting the docstring's claim that "ENVELOPE values stay authoritative when both representations are present." RFC 3501 permits an IMAP server to return `""` instead of `NIL` for an absent subject; a subject encoded as `=?UTF-8?Q??=` (empty encoded-word) could also decode to `""` through `decodeMIMEHeader`. In either case the guard fires and the ENVELOPE representation is replaced.

The fix is straightforward: use `header.subject == nil` (and `header.from == nil`) instead of `?.isEmpty != false`. For `from` the issue is additionally moot today because `applyEnvelope` never writes an empty string to `header.from`, but the nil check makes the guard's semantics unambiguous in both places.

Also noted (not a correctness issue): `EMLParser.buildMessageInfo(from: fields)` is called to share address/subject parsing, but its `parsed.date` output is discarded in favour of a direct `parseEnvelopeDate(rawDate)` call. The choice of parser is deliberate and consistent with the IMAP path, but the `parseRFC2822Date` work inside `buildMessageInfo` runs on every invocation for no benefit.

- **LOW** `Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift:144`

Sources/SwiftMail/IMAP/IMAP/Handler/FetchMessageInfoHandler.swift
jdudley: Fixed in 07cf3e47f9e3d46f6efcdd306f7ca0c35e957b06. Subject and from now backfill only when their ENVELOPE values are nil, so an explicit empty subject remains authoritative. Added regression coverage for that exact case; swift test passes 435 tests and strict SwiftLint reports 0 violations.

PR #207, opened 2026-08-06
Add read-only IMAP EXAMINE support

dariuscorvus:
## Summary

- add an `ExamineMailboxCommand` that emits IMAP `EXAMINE`
- expose `examineMailbox(_:)` on `IMAPServer` and `IMAPNamedConnection`
- capture tagged `READ-ONLY` / `READ-WRITE` response codes in `SelectHandler`
- cover command serialization and a real local IMAP integration path

This allows read-only clients to have the server enforce mailbox selection semantics instead of relying on discipline after `SELECT`.

## Verification

- `swift test` — 437 tests passed
- `git diff --check`

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPServer+Mailbox.swift
+++ b/Sources/SwiftMail/IMAP/IMAPServer+Mailbox.swift
@@ -44,6 +44,24 @@ extension IMAPServer {
     }
 
     /**
+     Select a mailbox in read-only mode
+
+     This method uses IMAP EXAMINE to make the mailbox current while asking the server
+     to enforce read-only access. STORE and EXPUNGE operations are not permitted while
+     the mailbox remains selected this way.
+
+     - Parameter mailboxName: The name of the mailbox to select read-only
+     - Returns: Status information about the selected mailbox
+     - Throws:
+     - `IMAPError.selectFailed` if the mailbox cannot be selected
+     - `IMAPError.connectionFailed` if not connected
+     */
+    @discardableResult public func examineMailbox(_ mailboxName: String) async throws -> Mailbox.Selection {
+        let command = ExamineMailboxCommand(mailboxName: resolveMailboxPath(mailboxName))
+        return try await executeCommand(command)
+    }
+
+    /**
      Close the currently selected mailbox
 
      This method closes the currently selected mailbox and expunges any messages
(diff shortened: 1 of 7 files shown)


Sources/SwiftMail/IMAP/IMAPServer+Mailbox.swift
```diff
@@ -43,6 +43,24 @@ extension IMAPServer {
...
+     the mailbox remains selected this way.
+
+     - Parameter mailboxName: The name of the mailbox to select read-only
+     - Returns: Status information about the selected mailbox
+     - Throws:
+     - `IMAPError.selectFailed` if the mailbox cannot be selected
+     - `IMAPError.connectionFailed` if not connected
+     */
+    @discardableResult public func examineMailbox(_ mailboxName: String) async throws -> Mailbox.Selection {
+        let command = ExamineMailboxCommand(mailboxName: resolveMailboxPath(mailboxName))
```
odrobnik, line 60: **MEDIUM** — After authenticating to a server with personal namespace `INBOX.`, disconnect and reconnect, then call `examineMailbox("Sent")`. Disconnect clears cached namespaces while retaining authentication. This line therefore constructs `EXAMINE "Sent"`; only afterward does `executeCommand` reauthenticate and learn that the required path is `INBOX.Sent`, so EXAMINE fails with NO/selectFailed. Authenticate before resolving, as the new `IMAPNamedConnection` overload does.

odrobnik (changes requested):
Requesting changes: the new primary-connection EXAMINE API resolves namespace-relative mailbox names before transparent reauthentication refreshes namespace metadata. The named-connection API uses the correct authenticate-before-resolve ordering.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer+Mailbox.swift:60`

Sources/SwiftMail/IMAP/IMAPServer+Mailbox.swift
odrobnik: Fixed by ensuring the primary connection is authenticated before resolving the EXAMINE mailbox path, so transparent reconnects refresh namespace metadata before constructing the command. Added a local integration regression covering disconnect/reconnect with the INBOX. personal namespace.

odrobnik:
MissionControl opened #211 from `agent/pr-207-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #211 merges. Thanks for the original contribution!

odrobnik:
merged via #211

Issue #194, opened 2026-07-26
Expose verified COPYUID mappings from IMAP COPY and MOVE


rajzik:
## Summary

Expose the server-provided `COPYUID` response from UID-based IMAP `COPY` and `MOVE` operations so callers can reliably correlate source UIDs with destination UIDs.

## Current behavior

In SwiftMail 1.8.0, the public copy and move APIs return `Void`. `CopyCommand` and `MoveCommand` also use `Void` as their result type, so a successful tagged response does not expose the `COPYUID` response code to callers.

The server may return:

```text
OK [COPYUID <destination UIDVALIDITY> <source UID set> <destination UID set>] ...
```

The source and destination UID sets encode an exact ordered mapping. That mapping is needed by clients that persist message identity or reconcile interrupted operations.

Relevant code:

- `Sources/SwiftMail/IMAP/IMAP/Commands/ServerCommands.swift`
- `Sources/SwiftMail/IMAP/IMAP/Commands/MoveCommand.swift`
- `Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift`
- `Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift`
- `Sources/SwiftMail/IMAP/IMAPNamedConnection+Manipulation.swift`

## Why this matters

An IMAP move implemented as copy followed by targeted source removal can be interrupted after the copy succeeds. Without the verified source-to-destination UID mapping, a client cannot safely determine whether to resume deletion, retry the copy, or migrate local identity to the destination message.

Using `Message-ID` matching is not a safe replacement because Message-IDs are not guaranteed to be unique or present. Retrying without reconciliation can create duplicates; deleting after an ambiguous match can remove the wrong message.

## Requested API behavior

Please return a public, `Sendable` transfer result for UID-based copy and move operations that represents:

- the destination mailbox UIDVALIDITY from `COPYUID`;
- the exact source-to-destination UID mapping;
- whether the server omitted `COPYUID`, represented distinctly from a verified empty mapping.

The concrete API shape is open. For example, it could be a result value returned by the existing methods or new result-returning overloads if source compatibility is preferred.

## Acceptance criteria

- UID `COPY` exposes the tagged response's `COPYUID` data through both `IMAPServer` and `IMAPNamedConnection`.
- UID `MOVE` exposes the same data when the server returns `COPYUID`.
[...]

PR #208, opened 2026-08-08
Add configurable MOVE fallback and preserve COPYUID evidence

tabmail-kmyi:
## What changed

- Expose case-insensitive MOVE capability detection on IMAPServer and IMAPNamedConnection.
- Add a MoveFallbackPolicy parameter to the existing move APIs on both surfaces. The default preserves COPY + STORE \Deleted + EXPUNGE fallback; the disabled policy requires MOVE and emits MOVE without requiring UIDPLUS.
- Return COPYUID mappings from COPY and MOVE, including the RFC 6851-recommended untagged OK response before EXPUNGE responses.
- Report malformed or conflicting COPYUID evidence with the typed malformedCopyUIDAfterTaggedOK error. That error explicitly means the provider command completed and must not be resent.

## Why

The fallback choice belongs on the existing move API so callers can require server MOVE without learning a second entry point. MOVE capability tokens are ASCII case-insensitive, and UIDPLUS controls optional mapping evidence rather than whether UID MOVE is legal.

A tagged OK is authoritative that COPY or MOVE completed. Keeping malformed mapping distinct from an omitted mapping lets durable callers reconcile or resync without repeating an already-committed mutation. Retaining untagged OK COPYUID also supports the response ordering recommended by RFC 6851.

## Compatibility

Existing callers keep the same source-compatible move signature and the same fallback behavior through the default copyStoreExpunge policy. Callers that require one MOVE command pass disabled. The separate moveAtomically proposal has been removed.

## Validation

- swift build passed.
- swift test: 457 tests in 59 suites passed.
- swiftlint lint --strict --no-cache: 0 violations across 288 files.
- Integration coverage verifies policy selection, server and named-connection conveniences, mixed-case MOVE capability spellings, and no fallback commands under disabled.
- Handler coverage verifies tagged and untagged COPYUID, malformed and conflicting evidence, and tagged failure precedence.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
@@ -20,12 +20,16 @@ final class MoveHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
         do {
             succeedWithResult(try extractCopyUID(from: response))
         } catch {
-            failWithError(error)
+            // The tagged OK is authority that MOVE completed. A malformed
+            // optional COPYUID makes only the address mapping unusable; turning
+            // it into command failure would make a durable caller retry an
+            // already-committed move.
+            succeedWithResult(nil)
         }
     }
 
     override func handleTaggedErrorResponse(_ response: TaggedResponse) {
-        failWithError(IMAPError.commandFailed("Move failed: \(String(describing: response.state))"))
+        failWithError(IMAPError.moveFailed(String(describing: response.state)))
     }
 }
 
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
@@ -64,7 +64,20 @@ final class MoveHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
     }
 
     override func handleTaggedErrorResponse(_ response: TaggedResponse) {
-        failWithError(IMAPError.moveFailed(String(describing: response.state)))
+        let reason = String(describing: response.state)
+        let partialCopyUID = lock.withLock {
+            malformedCopyUIDReason == nil ? retainedCopyUID : nil
+        }
+        if let partialCopyUID {
+            failWithError(
+                IMAPError.moveFailedAfterPartialCompletion(
+                    copyUID: partialCopyUID,
+                    reason: reason
+                )
+            )
+        } else {
+            failWithError(IMAPError.moveFailed(reason))
+        }
     }
 }
 
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
@@ -9,22 +9,57 @@ import NIOConcurrencyHelpers
 
 /// Handler for IMAP MOVE command.
 ///
-/// Extracts the `COPYUID` response code from the tagged OK when the server includes one
-/// (RFC 6851 §3.3). Returns `nil` when the server omits `COPYUID`.
+/// Retains a valid `COPYUID` from either the untagged `OK` recommended before EXPUNGE
+/// responses or the final tagged `OK` (RFC 6851 §§3.3, 4.3). Returns `nil` only when
+/// the successful response contains no `COPYUID` evidence.
 final class MoveHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @unchecked Sendable {
     typealias ResultType = CopyUID?
+
+    private var retainedCopyUID: CopyUID?
+    private var malformedCopyUIDReason: String?
+
+    override func handleUntaggedResponse(_ response: Response) -> Bool {
+        if super.handleUntaggedResponse(response) {
+            return true
+        }
+
+        guard case .untagged(.conditionalState(.ok(let text))) = response,
+              let code = text.code,
+              case .uidCopy(let data) = code
+        else {
+            return false
+        }
+
+        recordUntaggedCopyUID(data)
+        return false
+    }
 
     override func handleTaggedOKResponse(_ response: TaggedResponse) {
         super.handleTaggedOKResponse(response)
 
+        var taggedCopyUID: CopyUID?
+        var taggedMalformedReason: String?
         do {
-            succeedWithResult(try extractCopyUID(from: response))
+            taggedCopyUID = try extractCopyUID(from: response)
         } catch {
-            // The tagged OK is authority that MOVE completed. A malformed
-            // optional COPYUID makes only the address mapping unusable; turning
-            // it into command failure would make a durable caller retry an
-            // already-committed move.
-            succeedWithResult(nil)
+            taggedMalformedReason = String(describing: error)
+        }
+
+        let resolution = lock.withLock { () -> (copyUID: CopyUID?, malformedReason: String?) in
+            if let reason = malformedCopyUIDReason ?? taggedMalformedReason {
+                return (nil, reason)
+            }
+            if let retainedCopyUID, let taggedCopyUID,
+               !retainedCopyUID.isEquivalent(to: taggedCopyUID) {
+                return (nil, "Conflicting COPYUID mappings in untagged and tagged OK responses")
+            }
+            return (taggedCopyUID ?? retainedCopyUID, nil)
+        }
+
+        if let reason = resolution.malformedReason {
+            failWithError(IMAPError.malformedCopyUIDAfterTaggedOK(reason))
+        } else {
+            succeedWithResult(resolution.copyUID)
         }
     }
 
@@ -34,6 +69,26 @@ final class MoveHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
 }
 
 private extension MoveHandler {
+    func recordUntaggedCopyUID(_ data: NIOIMAPCore.ResponseCodeCopy) {
+        do {
+            let copyUID = try CopyUID(nio: data)
+            lock.withLock {
+                if let retainedCopyUID, !retainedCopyUID.isEquivalent(to: copyUID) {
+                    malformedCopyUIDReason =
+                        "Conflicting COPYUID mappings in multiple untagged OK responses"
+                } else if retainedCopyUID == nil {
+                    retainedCopyUID = copyUID
+                }
+            }
+        } catch {
+            lock.withLock {
+                if malformedCopyUIDReason == nil {
+                    malformedCopyUIDReason = String(describing: error)
+                }
+            }
+        }
+    }
+
     func extractCopyUID(from response: TaggedResponse) throws -> CopyUID? {
         guard case .ok(let text) = response.state,
               let code = text.code,
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
@@ -12,7 +12,10 @@ import NIOConcurrencyHelpers
 /// Retains a valid `COPYUID` from either the untagged `OK` recommended before EXPUNGE
 /// responses or the final tagged `OK` (RFC 6851 §§3.3, 4.3). Returns `nil` only when
 /// the successful response contains no `COPYUID` evidence.
-final class MoveHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @unchecked Sendable {
+final class MoveHandler:
+    BaseIMAPCommandHandler<CopyUID?>,
+    IMAPCommandHandler,
+    @unchecked Sendable {
     typealias ResultType = CopyUID?
 
     private var retainedCopyUID: CopyUID?
@@ -76,9 +79,10 @@ final class MoveHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
                 )
             )
         } else {
-            failWithError(IMAPError.moveFailed(reason))
+            failWithError(IMAPError.moveFailedAfterPossiblePartialCompletion(reason))
         }
     }
+
 }
 
 private extension MoveHandler {
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift
@@ -58,7 +58,11 @@ final class CopyHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
         do {
             succeedWithResult(try extractCopyUID(from: response))
         } catch {
-            failWithError(error)
+            // COPY itself succeeded; only its optional destination mapping was
+            // malformed. Returning nil prevents a durable caller from copying
+            // the same messages again while keeping tagged NO/BAD failures on
+            // the typed error path below.
+            succeedWithResult(nil)
         }
     }
 
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift
@@ -58,11 +58,7 @@ final class CopyHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
         do {
             succeedWithResult(try extractCopyUID(from: response))
         } catch {
-            // COPY itself succeeded; only its optional destination mapping was
-            // malformed. Returning nil prevents a durable caller from copying
-            // the same messages again while keeping tagged NO/BAD failures on
-            // the typed error path below.
-            succeedWithResult(nil)
+            failWithError(IMAPError.malformedCopyUIDAfterTaggedOK(String(describing: error)))
         }
     }
 
--- a/Sources/SwiftMail/IMAP/IMAPError.swift
+++ b/Sources/SwiftMail/IMAP/IMAPError.swift
@@ -22,11 +22,17 @@ public enum IMAPError: Error {
     case storeFailed(String)
     case expungeFailed(String)
     case moveFailed(String)
+    /// MOVE may have changed server state before failing, but no trustworthy mapping is available.
+    /// Callers must refresh both mailboxes and must not retry the original identifiers.
+    case moveFailedAfterPossiblePartialCompletion(String)
     /// MOVE completed for the verified subset in `copyUID` before ending with tagged NO/BAD.
     /// Callers should reconcile both mailboxes from this mapping and must not retry blindly.
     case moveFailedAfterPartialCompletion(copyUID: CopyUID, reason: String)
-    /// The command completed with tagged OK, but its present COPYUID evidence was malformed
-    /// or internally conflicting. The provider operation completed; callers must not resend it.
+    /// Fallback COPY completed and created the verified destinations in `copyUID`, but a later
+    /// STORE or EXPUNGE failed. Source flags/removal are unknown and must be refreshed.
+    case moveFallbackFailedAfterCopy(copyUID: CopyUID, reason: String)
+    /// The command completed with tagged OK, but its present COPYUID evidence was malformed,
+    /// conflicting, or unverifiable. The provider operation completed; callers must not resend it.
     case malformedCopyUIDAfterTaggedOK(String)
     case commandNotSupported(String)
     case authFailed(String)
@@ -75,10 +81,14 @@ extension IMAPError: CustomStringConvertible {
                 return "Expunge failed: \(reason)"
             case .moveFailed(let reason):
                 return "Move failed: \(reason)"
+            case .moveFailedAfterPossiblePartialCompletion(let reason):
+                return "Move may have partially completed before failing: \(reason)"
             case .moveFailedAfterPartialCompletion(_, let reason):
                 return "Move partially completed before failing: \(reason)"
+            case .moveFallbackFailedAfterCopy(_, let reason):
+                return "Move fallback copied messages before failing: \(reason)"
             case .malformedCopyUIDAfterTaggedOK(let reason):
-                return "Command completed but returned malformed COPYUID data: \(reason)"
+                return "Command completed but returned untrusted COPYUID data: \(reason)"
             case .commandNotSupported(let reason):
                 return "Command not supported: \(reason)"
             case .authFailed(let reason):
@@ -133,10 +143,14 @@ extension IMAPError: LocalizedError {
                 return "Failed to expunge deleted messages: \(reason)"
             case .moveFailed(let reason):
                 return "Failed to move messages: \(reason)"
+            case .moveFailedAfterPossiblePartialCompletion(let reason):
+                return "The server may have moved or copied messages before failing: \(reason)"
             case .moveFailedAfterPartialCompletion(_, let reason):
                 return "The server moved a verified subset of messages before failing: \(reason)"
+            case .moveFallbackFailedAfterCopy(_, let reason):
+                return "The fallback created verified destination copies, but source state is unknown: \(reason)"
             case .malformedCopyUIDAfterTaggedOK(let reason):
-                return "The command completed, but its COPYUID mapping was invalid: \(reason)"
+                return "The command completed, but its COPYUID mapping could not be trusted: \(reason)"
             case .commandNotSupported(let reason):
                 return "The requested command is not supported by the server: \(reason)"
             case .authFailed(let reason):
@@ -170,10 +184,24 @@ extension IMAPError: LocalizedError {
                 return "Check that your email provider supports XOAUTH2 for IMAP connections."
             case .moveFailedAfterPartialCompletion:
                 return "Do not retry blindly. Reconcile both mailboxes using the verified COPYUID mapping."
+            case .moveFallbackFailedAfterCopy:
+                return "Do not retry. Reconcile the destination copies using COPYUID and refresh the source mailbox."
+            case .moveFailed, .moveFailedAfterPossiblePartialCompletion:
+                return "Do not retry. Refresh both mailboxes and reconcile their current state first."
             case .malformedCopyUIDAfterTaggedOK:
                 return "Do not retry. The command completed; refresh the affected mailboxes before continuing."
             default:
                 return "Check the error details and try again."
         }
+    }
+}
+
+extension IMAPError {
+    static func moveFallbackFailed(after copyUID: CopyUID?, underlying error: Error) -> IMAPError {
+        let reason = "COPY completed before fallback failed: \(error)"
+        if let copyUID {
+            return .moveFallbackFailedAfterCopy(copyUID: copyUID, reason: reason)
+        }
+        return .moveFailedAfterPossiblePartialCompletion(reason)
     }
 }
--- a/Sources/SwiftMail/IMAP/IMAPError.swift
+++ b/Sources/SwiftMail/IMAP/IMAPError.swift
@@ -22,6 +22,9 @@ public enum IMAPError: Error {
     case storeFailed(String)
     case expungeFailed(String)
     case moveFailed(String)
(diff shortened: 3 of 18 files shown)


odrobnik:
if we have a move function that only does the move without fallback, couldn’t we join that with the regular move function but have an option to use fallback or not.

Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift
```diff
@@ -58,7 +58,11 @@ final class CopyHandler: BaseIMAPCommandHandler<CopyUID?>, IMAPCommandHandler, @
         do {
             succeedWithResult(try extractCopyUID(from: response))
         } catch {
-            failWithError(error)
+            // COPY itself succeeded; only its optional destination mapping was
+            // malformed. Returning nil prevents a durable caller from copying
+            // the same messages again while keeping tagged NO/BAD failures on
+            // the typed error path below.
+            succeedWithResult(nil)
```
odrobnik: **MEDIUM** — If a server replies `A001 OK [COPYUID 1 1:2 200] COPY completed`, `CopyUID(nio:)` detects the cardinality mismatch, but this catch returns `nil`. The caller receives exactly the same value as when COPYUID was absent, so it cannot detect or quarantine a corrupt mapping despite issue #194 requiring malformed mappings to fail explicitly; `MoveHandler` now has the same regression.

Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -96,6 +96,15 @@ public actor IMAPServer {
         capabilities.contains(.uidPlus)
     }
 
+    /// Whether the primary connection advertised MOVE (RFC 6851).
+    ///
+    /// This reports the advertised capability only. ``move(messages:to:)`` issues a real `MOVE`
+    /// for a UID-based set only when the server also advertises UIDPLUS, and otherwise falls back
+    /// to COPY + STORE `\Deleted` + EXPUNGE.
+    public var supportsMove: Bool {
+        capabilities.contains(.move)
```
odrobnik: **MEDIUM** — A standards-compliant server may advertise `move` or `Move` because RFC 6851 makes protocol tokens case-insensitive. NIOIMAPCore 0.3.0 preserves that raw spelling and synthesizes case-sensitive `Capability` equality, so `contains(.move)` is false: `supportsMove` reports false and `moveAtomically` throws `commandNotSupported` without issuing the supported command. The named-connection accessor has the same failure.

odrobnik (changes requested):
Requesting changes for three correctness/API-contract issues:

- The new handler behavior collapses malformed `COPYUID` data into the same `nil` used for an omitted response code. That reverses issue #194's explicit malformed-mapping contract and contradicts the public API documentation.
- `supportsMove` uses case-sensitive `Capability` equality even though RFC 6851 capability tokens are case-insensitive; the pinned NIOIMAPCore parser preserves the server's spelling.
- `MoveHandler` only examines the tagged completion and therefore discards a valid `COPYUID` sent in the RFC 6851-recommended untagged `OK` before EXPUNGE responses.

These affect both `IMAPServer` and `IMAPNamedConnection` where noted.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift:65`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer.swift:105`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift` — A UIDPLUS server following RFC 6851 section 4.3 can reply `* OK [COPYUID 42 1 101]`, then EXPUNGE responses, then `A001 OK Done`. The base handler buffers the untagged OK, but `extractCopyUID` inspects only the final tagged response, so both move APIs return `nil` even though the server supplied the verified mapping needed for reconciliation.

Sources/SwiftMail/IMAP/IMAP/Handler/ServerHandlers.swift
tabmail-kmyi: Fixed in 6ac6b84. A present malformed COPYUID now throws the typed malformedCopyUIDAfterTaggedOK error instead of collapsing to nil; omission remains nil. Focused handler tests and the full 457-test suite pass.

Sources/SwiftMail/IMAP/IMAPServer.swift
tabmail-kmyi: Fixed in 6ac6b84. MOVE capability matching now uses ASCII-insensitive token comparison on both IMAPServer and IMAPNamedConnection, with mixed-case parser coverage. The full 457-test suite passes.

tabmail-kmyi:
Updated in `6ac6b84`.

- The separate atomic-only entry point is gone. Existing `move` APIs now take a defaulted `MoveFallbackPolicy`; current callers retain COPY/STORE/EXPUNGE fallback, while callers that require one server MOVE pass `.disabled`.
- `MoveHandler` now retains valid `COPYUID` evidence from the RFC 6851 untagged `OK` before EXPUNGE responses, reconciles tagged/untagged evidence, and rejects malformed or conflicting evidence with the typed tagged-OK completion error.
- The two inline findings are also fixed and resolved.

Validation: `swift build` passed; all 457 tests in 59 suites passed; strict SwiftLint reported 0 violations across 288 files.

Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
```diff
@@ -9,27 +9,86 @@ import NIOConcurrencyHelpers
...
+        if let reason = resolution.malformedReason {
+            failWithError(IMAPError.malformedCopyUIDAfterTaggedOK(reason))
+        } else {
+            succeedWithResult(resolution.copyUID)
         }
     }
 
     override func handleTaggedErrorResponse(_ response: TaggedResponse) {
-        failWithError(IMAPError.commandFailed("Move failed: \(String(describing: response.state))"))
+        failWithError(IMAPError.moveFailed(String(describing: response.state)))
```
odrobnik: **MEDIUM** — RFC 6851 §§3.3 and 4.3 permit a MOVE to move only a subset, emit an untagged OK with COPYUID for the messages that did move, and then finish with a tagged NO. The handler now retains that valid mapping, but this error path discards it and exposes only moveFailed. A durable caller therefore cannot reconcile the UIDs that were already moved, contrary to issue #194's mapping requirement, and may retry or resync without the evidence the server supplied.

Sources/SwiftMail/IMAP/Models/CopyUID.swift
```diff
@@ -4,10 +4,11 @@ import NIOIMAPCore
...
 ///
 /// When the server supports UIDPLUS (RFC 4315) and returns a `COPYUID` response code in the
-/// tagged `OK`, this value carries the exact source-to-destination UID mapping in the order
-/// the server provided it. When `COPYUID` is absent — either because the server doesn't
-/// advertise UIDPLUS or chose not to include the code — the corresponding method returns
-/// `nil` instead.
+/// successful `OK` response, this value carries the exact source-to-destination UID mapping
+/// in the order the server provided it. When `COPYUID` is absent — either because the server
+/// doesn't advertise UIDPLUS or chose not to include the code — the corresponding method
+/// returns `nil` instead. Malformed or conflicting evidence throws
```
odrobnik: **MEDIUM** — A server can reply with tagged OK and `[COPYUID 42 1,1 101,102]`. The pinned parser preserves those repeated range elements, and CopyUID.expand produces two source entries and two destination entries, so the count-only guard accepts and returns mappings `(1,101)` and `(1,102)`. Semantically the source UID set has only one member, so this is malformed/contradictory evidence that the newly documented contract says must throw rather than be exposed as verified.

Sources/SwiftMail/IMAP/IMAPError.swift
```diff
@@ -123,6 +128,8 @@ extension IMAPError: LocalizedError {
                 return "Failed to expunge deleted messages: \(reason)"
             case .moveFailed(let reason):
                 return "Failed to move messages: \(reason)"
+            case .malformedCopyUIDAfterTaggedOK(let reason):
```
odrobnik, line 152: **MEDIUM** — For malformedCopyUIDAfterTaggedOK, LocalizedError.recoverySuggestion falls through to `Check the error details and try again.` An app or error UI following that public recovery guidance will retry a COPY or MOVE that the new case explicitly says completed and must not be resent, potentially creating duplicates or acting on the wrong post-MOVE state.

Sources/SwiftMail/IMAP/IMAPNamedConnection+Manipulation.swift
```diff
@@ -45,23 +48,40 @@ extension IMAPNamedConnection {
...
     @discardableResult
     public func move<T: MessageIdentifier>(
         messages identifierSet: MessageIdentifierSet<T>,
-        to destinationMailbox: String
+        to destinationMailbox: String,
+        fallback: MoveFallbackPolicy = .copyStoreExpunge
     ) async throws -> CopyUID? {
-        if capabilities.contains(.move) && (T.self != UID.self || capabilities.contains(.uidPlus)) {
+        if case .disabled = fallback {
+            guard capabilities.containsMoveCapability else {
```
odrobnik, line 94: **MEDIUM** — After a recyclable connection failure, disconnectBody clears the capability snapshot while the named handle retains its authentication closure. If re-authentication/SELECT is in flight against a server that advertises MOVE only after login, a concurrent disabled-policy move reaches this guard before executeCommand can call ensureAuthenticated and incorrectly throws commandNotSupported instead of waiting behind authentication and issuing MOVE. The mirrored IMAPServer guard at line 42 has the same pre-authentication TOCTOU.

Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift
```diff
@@ -18,62 +20,84 @@ extension IMAPServer {
...
+     - ``IMAPError/malformedCopyUIDAfterTaggedOK(_:)`` after a successful command with
+       malformed or conflicting COPYUID evidence; the command completed and must not be resent
      - Note: Logs move operations at info level with message count and destination
      */
     @discardableResult
     public func move<T: MessageIdentifier>(
         messages identifierSet: MessageIdentifierSet<T>,
-        to destinationMailbox: String
+        to destinationMailbox: String,
+        fallback: MoveFallbackPolicy = .copyStoreExpunge
```
odrobnik: **LOW** — Replacing the existing two-parameter public method with a three-parameter method whose last argument has a default preserves direct call syntax, but not the method's function type. Existing clients that store or pass `move(messages:to:)` as a two-argument async function value no longer compile because default arguments are not carried by function values. The single/header and IMAPNamedConnection overloads were replaced the same way, contradicting the PR's source-compatibility claim.

odrobnik (changes requested):
Requesting changes for five remaining correctness/API-contract issues. The happy-path fixes for mixed-case MOVE and tagged/untagged OK reconciliation are sound, but RFC-permitted partial-failure evidence is still lost; malformed UID sets can still be exposed as verified mappings; the new no-retry error advertises a retry through LocalizedError; the disabled-policy gate bypasses re-authentication on both connection surfaces; and the signature replacement is not fully source-compatible.

One review with 5 findings, each detailed in an inline comment:

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift:67`
- **MEDIUM** `Sources/SwiftMail/IMAP/Models/CopyUID.swift:10`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPError.swift:131`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPNamedConnection+Manipulation.swift:70`
- **LOW** `Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift:39`

tabmail-kmyi:
While validating the review fixes in release configuration, I confirmed that `swift test -c release` also fails on untouched `main`: the response-buffer tests from #179 are unconditional, but their internal probes are compiled only under `#if DEBUG`.

I opened #210 as a deliberately separate, four-line prerequisite fix. It keeps those internal read-only probes available in release builds, so the existing tests run rather than being skipped; both full debug and release suites pass (435 tests each).

I would prefer to land #210 first, then update this branch from the green `main` baseline before posting the full #208 revision. That keeps this PR focused and avoids hand-waving the release failure as pre-existing.

Sources/SwiftMail/IMAP/IMAPError.swift
tabmail-kmyi: Fixed in f0d7686. malformedCopyUIDAfterTaggedOK now has explicit Do not retry recovery guidance that directs callers to refresh affected mailboxes. A focused recovery test locks this behavior.

Sources/SwiftMail/IMAP/Models/CopyUID.swift
tabmail-kmyi: Fixed in f0d7686. COPYUID expansion now rejects repeated or overlapping source and destination UIDs before constructing a mapping. Parameterized tests cover singleton and range overlaps for both COPY and MOVE.

Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift
tabmail-kmyi: Fixed in f0d7686. All five original two-argument overloads are restored as exact wrappers, while the new three-argument policy overloads have no default. Compile-time function-value assignments cover server message-set, single, header, and both named forms.

Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
tabmail-kmyi: Fixed in f0d7686. Tagged NO/BAD after valid untagged COPYUID now throws moveFailedAfterPartialCompletion with the exact verified mapping and non-retry reconciliation guidance. A focused handler test covers the partial mapping and recovery contract.

Sources/SwiftMail/IMAP/IMAPNamedConnection+Manipulation.swift
tabmail-kmyi: Fixed in f0d7686. Server and named policy-bearing MOVE paths now authenticate and reconnect before taking the capability snapshot. The regression test disconnects both surfaces and proves UID MOVE is emitted after reauthentication instead of a stale commandNotSupported rejection.

tabmail-kmyi:
Updated in f0d7686 after prerequisite #210 landed. All five requested fixes are now addressed: verified partial COPYUID evidence survives tagged MOVE failure; repeated and overlapping source/destination UID evidence is rejected; completed-operation recovery no longer suggests retry; server and named MOVE capability decisions occur after reauthentication; and all five original two-argument overloads are restored alongside explicit policy overloads. Validation: 462 debug tests, 462 release tests on the #210 base, optimized Swift build, generic iOS build, strict SwiftLint, and two consecutive focused review rounds. This revision remains limited to eight existing files (+250/-19); adjacent lifecycle hardening was deliberately left for separate one-item PRs.

Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift
```diff
@@ -18,62 +32,115 @@ extension IMAPServer {
...
-        } else {
-            // Fall back to COPY + DELETE + targeted expunge when UIDPLUS is available.
-            let copyUID = try await copy(messages: identifierSet, to: destinationMailbox)
-            try await store(flags: [.deleted], on: identifierSet, operation: .add)
-            try await expungeMoveFallback(messages: identifierSet)
-            return copyUID
         }
+
+        if capabilities.containsMoveCapability
+            && (T.self != UID.self || capabilities.contains(.uidPlus)) {
```
odrobnik: **HIGH** — UIDPLUS capability matching is still case-sensitive here (and in the named-connection mirror), even though IMAP capability names are case-insensitive and the pinned parser preserves server spelling. If a server advertises `MOVE` and lowercase `uidplus`, a default UID move incorrectly enters the COPY/STORE/EXPUNGE fallback. The fallback's UIDPLUS check fails for the same reason, so it emits plain EXPUNGE; any unrelated message already marked `\Deleted` is then permanently removed even though the server supported both atomic MOVE and targeted UID EXPUNGE.

Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
```diff
@@ -9,27 +9,99 @@ import NIOConcurrencyHelpers
...
+        }
+        if let partialCopyUID {
+            failWithError(
+                IMAPError.moveFailedAfterPartialCompletion(
+                    copyUID: partialCopyUID,
+                    reason: reason
+                )
+            )
+        } else {
+            failWithError(IMAPError.moveFailed(reason))
```
odrobnik: **HIGH** — Absence of a valid COPYUID mapping does not mean that a tagged NO/BAD moved nothing: RFC 6851 explicitly permits MOVE to fail after moving a subset, including on servers without UIDPLUS. This branch returns ordinary `moveFailed`, whose public recovery suggestion tells callers to try again. For a sequence-number MOVE, the server can move and EXPUNGE sequence 1, renumber the mailbox, then return NO; retrying the original `1:2` set can move a different message that was never requested. The no-mapping failure therefore needs an unknown/possible-partial-completion, non-retry contract rather than the retryable failure.

Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift
```diff
@@ -18,62 +32,115 @@ extension IMAPServer {
...
+            return try await executeMove(messages: identifierSet, to: destinationMailbox)
+        }
+
+        // Preserve the existing fallback, including targeted UID EXPUNGE when UIDPLUS is available.
+        let copyUID = try await copy(messages: identifierSet, to: destinationMailbox)
+        try await store(flags: [.deleted], on: identifierSet, operation: .add)
```
odrobnik: **MEDIUM** — The fallback keeps COPY's verified mapping only in a local that is returned after STORE and EXPUNGE both succeed. If `UID COPY 1 Archive` succeeds with `[COPYUID 42 1 101]` and the subsequent STORE, EXPUNGE, or connection fails, this line throws an unrelated error and discards the mapping; the named-connection path does the same. A durable caller cannot reconcile the already-created destination UID and may retry the whole move, creating a duplicate—the exact partial-success case issue #194 requires the mapping to make recoverable.

Sources/SwiftMail/IMAP/Models/CopyUID.swift
```diff
@@ -65,9 +70,24 @@ extension CopyUID {
                 throw IMAPError.commandFailed("COPYUID expansion exceeds \(maxExpandedUIDs) UIDs")
             }
             for raw in lower...upper {
+                guard seen.insert(raw).inserted else {
```
odrobnik, line 82: **MEDIUM** — The new malformed-set validation rejects duplicates but still accepts the RFC 4315-forbidden `*` member. The pinned NIOIMAPCore parser maps `*` to the maximum raw identifier, so `[COPYUID 42 * 101]` reaches this loop as a unique singleton and is exposed as a verified mapping from UID 4294967295 to 101. Callers can then migrate identity using a source UID the response was not allowed to contain instead of receiving `malformedCopyUIDAfterTaggedOK`.

Sources/SwiftMail/IMAP/Models/CopyUID.swift
odrobnik: **MEDIUM** — The documented malformed-evidence guarantee is still not enforced against the command's requested UID set. If the client issues `UID COPY 1 Archive` (or UID MOVE 1) and the server returns `[COPYUID 42 2 101]`, the source and destination sets are unique and have equal cardinality, so this implementation exposes `(2, 101)` as verified even though UID 2 was extraneous and could not have been copied by that command. RFC 4315 forbids extraneous UIDs; accepting this mapping can make a caller update or delete identity for the wrong message. The handler/command must validate response sources are a subset of the requested UID set.

odrobnik (changes requested):
Requesting changes for five remaining correctness and safety issues. Case-insensitive UIDPLUS advertisements can still route a supported UID MOVE through COPY/STORE and an unscoped EXPUNGE; a tagged MOVE failure without COPYUID is still exposed as retryable even though MOVE may have partially completed; fallback failures discard a verified COPYUID mapping; and the malformed-evidence checks still accept RFC-forbidden wildcard and extraneous source UIDs. These can cause duplicate copies, loss of reconciliation evidence, or movement/deletion of messages outside the caller's intended set.

One review with 5 findings, each detailed in an inline comment:

- **HIGH** `Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift:66`
- **HIGH** `Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift:79`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift:72`
- **MEDIUM** `Sources/SwiftMail/IMAP/Models/CopyUID.swift:73`
- **MEDIUM** `Sources/SwiftMail/IMAP/Models/CopyUID.swift:10`

Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift
tabmail-kmyi: Fixed in c19836a. UIDPLUS now uses the same ASCII-insensitive bare-capability matcher as MOVE on both server and named surfaces. Lowercase uidplus coverage proves default UID MOVE stays atomic and fallback emits targeted UID EXPUNGE, never plain EXPUNGE.

Sources/SwiftMail/IMAP/IMAP/Handler/MoveHandler.swift
tabmail-kmyi: Fixed in c19836a. Tagged MOVE NO/BAD without trustworthy COPYUID now throws moveFailedAfterPossiblePartialCompletion with explicit do-not-retry reconciliation guidance. The focused handler test covers the no-mapping failure contract.

Sources/SwiftMail/IMAP/IMAPServer+Manipulation.swift
tabmail-kmyi: Fixed in c19836a. After fallback COPY succeeds, later STORE or EXPUNGE failure now preserves verified destination mapping in moveFallbackFailedAfterCopy; absent mapping becomes a non-retryable possible-partial failure. Server and named regressions cover this.

Sources/SwiftMail/IMAP/Models/CopyUID.swift
tabmail-kmyi: Fixed in c19836a. COPYUID expansion now rejects NIOIMAPCore maximum identifiers because the pinned parser makes wire wildcard and numeric 4294967295 indistinguishable. COPY and MOVE tests cover source, destination, and range wildcard forms plus the conservative numeric-maximum case.

Sources/SwiftMail/IMAP/Models/CopyUID.swift
tabmail-kmyi: Fixed in c19836a. COPY and MOVE commands now validate successful and partial COPYUID source evidence against the requested UID set before exposing it. Extraneous sources are rejected, and wildcard-bearing requests fail closed when provenance cannot be proven; focused and both-surface tests cover the contract.

tabmail-kmyi:
Implemented the five requested safety fixes in c19836a; the final branch head is 7aee922 after merging current upstream changes from #210 and #211.\n\n- UIDPLUS and MOVE capability matching is ASCII case-insensitive on both server surfaces.\n- Tagged MOVE NO/BAD without mapping is reported as possible partial completion with non-retry guidance.\n- COPY fallback failures after a successful COPY retain trusted COPYUID evidence, or report unknown partial completion when none is available.\n- COPYUID wildcard/maximum-UID ambiguity is rejected as untrusted evidence.\n- COPYUID source UIDs are validated against the requested UID set, including fail-closed handling when wildcard provenance cannot be proven.\n\nThe change remains scoped to these review findings; generic transport and command-handler hardening was deliberately excluded.\n\nFinal validation on 7aee922:\n- 475 debug tests in 61 suites passed.\n- 475 release tests in 61 suites passed.\n- Optimized Swift build passed.\n- Strict SwiftLint and git diff checks passed.\n- GitHub CI is green across macOS, iOS, Linux, Android, and lint.\n- Two consecutive three-reviewer targeted integrity rounds on the final head were clean.

odrobnik (approved):
Adversarial review of head 7aee922 — no findings.

**Capability matching** (`Capability+Move.swift`): `containsBareCapability` uses `(candidate | 0x20) == expected` with lowercase-only `expectedName` strings. This correctly maps any ASCII-cased server spelling to the expected lowercase byte for both MOVE and UIDPLUS. The `capability.value == nil` guard correctly excludes parameterized capabilities like `AUTH=PLAIN`. Applied consistently on both `IMAPServer` and `IMAPNamedConnection` surfaces.

**MoveHandler** (`MoveHandler.swift`): `handleUntaggedResponse` extracts the untagged COPYUID and returns `false` — the handler stays active until the tagged response arrives. `handleTaggedOKResponse` calls `super.handleTaggedOKResponse` first; for `ResultType = CopyUID?`, the cast `() as Any as? CopyUID?` fails cleanly (boxed Void is not Optional&lt;CopyUID&gt;), so `succeedWithResult` is never called from the base and `isCompleted` remains false. The subclass then calls `succeedWithResult` or `failWithError` exactly once. The lock (`NSRecursiveLock`) is acquired consistently in `recordUntaggedCopyUID` and in `handleTaggedOKResponse`'s resolution block; no re-entrancy issue because both code paths run on the NIO event loop. `handleTaggedErrorResponse` correctly reads `malformedCopyUIDReason` and `retainedCopyUID` under the same lock and maps to the right error type without calling super (which would conflict).

**COPYUID validation** (`CopyUID.swift`, `MoveCommand.swift`, `ServerCommands.swift`): The `UInt32.max` guard correctly fails closed on the wire wildcard / numeric maximum ambiguity. The per-set `seen: Set<UInt32>` correctly catches intra-set duplicates (source vs source, destination vs destination) while legitimately allowing the same UID to appear in both source and destination sets (different mailboxes). `sourceValidationFailure` skips validation for sequence-number commands (`T.self != UID.self`), fails closed when the request contains `UID.latest` (= `UInt32.max`, the unresolved `*`), and otherwise checks each mapping source against the requested set. `validate(copyUID:)` is applied after `executeCommand` for both COPY and MOVE; `validate(error:)` degrades `moveFailedAfterPartialCompletion` with bad sources to `moveFailedAfterPossiblePartialCompletion`, leaving all other error types unchanged.

**Fallback path** (`IMAPServer+Manipulation.swift`, `IMAPNamedConnection+Manipulation.swift`): `copy()` validates COPYUID before returning it. If STORE or EXPUNGE subsequently fails, the validated `copyUID` value (or nil when the server omitted COPYUID) is preserved in `moveFallbackFailedAfterCopy` or `moveFailedAfterPossiblePartialCompletion` respectively via `moveFallbackFailed(after:underlying:)`. The inner `do/catch` only covers STORE/EXPUNGE, so a COPY validation failure propagates directly without hitting that catch.

**Recovery guidance** (`IMAPError.swift`): Every new error case has an explicit "Do not retry" prefix in `recoverySuggestion`; the `default` fall-through to "Check the error details and try again" is not reachable by any of the new cases.

**Public API compatibility** (`IMAPServer+Manipulation.swift`, `IMAPNamedConnection+Manipulation.swift`): The two-argument overloads (`move(messages:to:)`, `move(message:to:)`, `move(header:to:)`) are independent functions wrapping the three-argument form — not defaults on the three-argument signature — so function-value assignments to `(Set, String) async throws -> CopyUID?` compile correctly.

**Reauthentication test** (`AtomicMoveTests.swift`): After `disconnect()`, `capabilities` is cleared to `[]`. `ensurePrimaryConnectionAuthenticated()` reconnects and logs in, refreshing capabilities (including MOVE). The MOVE command is then issued. The test server enforces `guard selectedMailbox != nil` for all UID commands; since reauthentication does not re-SELECT a mailbox, the server returns `NO No mailbox selected`, the handler produces `moveFailedAfterPossiblePartialCompletion`, and the commandLog still contains a `UID MOVE` entry — proving capability detection happened post-authentication rather than throwing stale `commandNotSupported`. The test logic is correct.

👍 No findings.

PR #209, opened 2026-08-08
Make SMTP submission timeouts configurable and RFC-aligned

jdudley:
## Summary

- add public `SMTPSubmissionTimeouts` configuration with RFC 5321 section 4.5.3.2 defaults
- send DATA content as sequential bounded 64 KiB buffers and restart the upload deadline after each successful flush
- give message-content upload and the final acceptance reply independent timeout budgets
- preserve the expired timeout stage in typed public diagnostics
- start the final-response timer only after the final content buffer flushes, so upload time cannot consume the server-processing allowance
- preserve typed `SMTPSendError` ambiguity and cancellation behavior for unsafe-to-retry outcomes
- document timeout customization and stage-specific diagnostics in the SMTP getting-started guide

The previous `SendContentCommand` path used one fixed 10-second timer that began before the full message was flushed and remained active while waiting for the final `250`. On slower or attachment-bearing sends, that could produce a spurious ambiguous timeout after the server had accepted the message, creating duplicate-delivery risk if a caller retried.

The revised path follows RFC 5321's per-buffer DATA timeout model: total upload allowance scales with a progressing message, while a single stalled buffer still fails with an ambiguous, unsafe-to-retry outcome.

## Validation

- `swift test` — 439 tests in 57 suites passed
- `swift test --filter SMTPTests` — 98 tests in 3 suites passed
- `swift build`
- `swiftlint lint --strict` — 0 violations

New integration coverage verifies a stalled 8 MiB content upload, a progressing 8 MiB upload whose total duration exceeds its per-buffer budget, and a delayed final `250` that arrives after the upload budget but within its fresh response budget.


Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
```diff
@@ -0,0 +1,128 @@
...
+        let completionPromise = channel.eventLoop.makePromise(of: Void.self)
+        let timeoutTask = channel.eventLoop.scheduleTask(
+            in: Self.nioTimeAmount(seconds: timeout)
+        ) {
+            if resolution.claim() {
+                completionPromise.fail(SMTPSubmissionTimeoutError())
+            }
+        }
+
+        channel.writeAndFlush(buffer).whenComplete { result in
```
odrobnik: **MEDIUM** — `SendContentCommand` is materialized as one `ByteBuffer`, and this single `writeAndFlush` is guarded by one absolute `contentUpload` deadline. RFC 5321 §4.5.3.2 requires the 3-minute DATA timeout per transfer buffer/TCP SEND so total allowance is proportional to message size. For example, a valid 25 MiB message over a healthy 1 Mbit/s uplink takes about 200 seconds: bytes can make continuous progress, yet the default timer fires at 180 seconds, closes the session, and returns an ambiguous/unsafe-to-retry failure (with duplicate-delivery risk if retried). Split the content into bounded chunks and reset the upload deadline after each successful chunk/progress event instead of timing the whole message as one operation.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
```diff
@@ -0,0 +1,128 @@
...
+        through channel: Channel,
+        timeout: TimeInterval
+    ) async throws {
+        let resolution = SMTPSubmissionWriteResolution()
+        let completionPromise = channel.eventLoop.makePromise(of: Void.self)
+        let timeoutTask = channel.eventLoop.scheduleTask(
+            in: Self.nioTimeAmount(seconds: timeout)
+        ) {
+            if resolution.claim() {
+                completionPromise.fail(SMTPSubmissionTimeoutError())
```
odrobnik: **LOW** — When the write budget expires here, `SMTPSubmissionTimeoutError` is classified as `SMTPSendError.Reason.timedOut`, whose public documentation and localized description say the client timed out waiting for the server's reply. During a stalled upload no final reply is being awaited. A peer that stops reading therefore produces a misleading response-timeout diagnostic, and callers/telemetry cannot tell whether `contentUpload` or `contentResponse` expired, leading operators to tune the wrong budget. Preserve the timeout stage in the marker/public reason, or make the public diagnostic accurately cover upload expiration.

odrobnik (changes requested):
The split final-response budget, cancellation bridge, and ambiguous-outcome teardown are coherent, but two issues remain. Most importantly, the 3-minute DATA upload timeout is applied once to the entire monolithic message buffer; [RFC 5321 §4.5.3.2](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.5.3.2) requires a timer for each data-transfer buffer so the overall allowance scales with message size. A continuously progressing large upload can therefore still be aborted and become unsafe to retry. The upload-expiry path also reuses a response-timeout marker, so its public diagnostic falsely says the client was waiting for a server reply. This was a read-only review; builds and tests were not run as requested.

- **MEDIUM** `Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift:107`
- **LOW** `Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift:103`

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
jdudley: Fixed in 996e63a. SMTPSubmissionTimeoutError now carries a public SMTPSendError.TimeoutStage, and Reason.timedOut preserves it. Upload expiration reports contentUpload and timed out uploading message content; final-reply expiration reports contentResponse separately.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
jdudley: Fixed in 996e63a. DATA is now emitted as sequential bounded 64 KiB buffers; each successful writeAndFlush starts a fresh contentUpload deadline. Added an integration test whose total upload duration exceeds the per-buffer budget while every progressing buffer completes in time.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
```diff
@@ -0,0 +1,171 @@
...
+                capacity: contentCount + (isFinalBuffer ? 2 : 0)
+            )
+            if contentCount > 0 {
+                buffer.writeBytes(data[index..<nextIndex])
+            }
+            if isFinalBuffer {
+                buffer.writeBytes([0x0D, 0x0A]) // CRLF
+            }
+
+            try await flushSubmissionBuffer(
```
odrobnik, line 176: **HIGH** — Awaiting every 64 KiB flush reopens the SMTPServer actor without any transaction gate. During a large DATA upload, a concurrent sendEmail can write `MAIL FROM` between content chunks; the server treats it as body bytes and may return 250 for a silently corrupted first message, so that caller reports false success. Serialize the whole connection transaction or otherwise prevent writes from interleaving across the chunk sequence.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
```diff
@@ -0,0 +1,171 @@
...
+    private func flushSubmissionData(
+        _ data: Data,
+        through channel: Channel,
+        timeout: TimeInterval,
+        timeoutStage: SMTPSendError.TimeoutStage
+    ) async throws {
+        var index = data.startIndex
+
+        repeat {
+            try Task.checkCancellation()
```
odrobnik, line 161: **MEDIUM** — On the first content iteration this check can observe cancellation after the caller's safe pre-content check but before any writeAndFlush. The outer catch nevertheless reports `.ambiguous` and `.unsafeToRetry`; an outbox can hold a message for manual review even though no content byte was sent and retrying is safe. Track whether the first buffer was dispatched and preserve the pre-content classification.

Sources/SwiftMail/SMTP/SMTPSendError.swift
```diff
@@ -85,8 +100,8 @@ public struct SMTPSendError: Error, Sendable, Equatable {
         /// The surrounding task was cancelled.
         case cancelled
 
-        /// No reply arrived within the command timeout.
-        case timedOut
+        /// One submission operation exceeded its stage-specific timeout.
+        case timedOut(TimeoutStage)
```
odrobnik: **MEDIUM** — Version 1.10.0 shipped public no-payload `Reason.timedOut`, including a public initializer intended for fabricated outcomes. Existing clients using `reason == .timedOut` or `reason: .timedOut` stop compiling after a 1.x update because the case now requires TimeoutStage. Add the stage through a source-compatible API or reserve this change for a major version.

Sources/SwiftMail/SMTP/SMTPSubmissionTimeouts.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// Maximum time to wait for the reply to `MAIL FROM`.
+    public let mailFromResponse: TimeInterval
+
+    /// Maximum time to wait for the reply to each `RCPT TO`.
+    public let recipientResponse: TimeInterval
+
+    /// Maximum time to wait for the reply to `DATA`.
+    public let dataResponse: TimeInterval
+
+    /// Maximum time to flush the message content and terminating dot.
```
odrobnik: **LOW** — The public comment promises an upload-wide maximum, but the implementation resets this value for every 64 KiB buffer. With `contentUpload: 30`, 100 buffers can each take 29 seconds and the call can run about 50 minutes instead of the documented 30-second ceiling. Document the value explicitly as a per-buffer maximum, including in the guide.

odrobnik (changes requested):
Requesting changes. The prior timeout findings are addressed, but per-buffer flushing now permits actor reentrancy to splice another SMTP operation into an in-flight DATA body, and first-buffer cancellation can be misclassified as unsafe to retry before any write. The timeout-stage change also breaks the shipped 1.10 public API, and `contentUpload` documents whole-upload rather than per-buffer semantics. Read-only review; builds/tests were not run as requested.

One review with 4 findings, each detailed in an inline comment:

- **HIGH** `Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift:124`
- **MEDIUM** `Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift:107`
- **MEDIUM** `Sources/SwiftMail/SMTP/SMTPSendError.swift:104`
- **LOW** `Sources/SwiftMail/SMTP/SMTPSubmissionTimeouts.swift:22`

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
jdudley: Fixed in d9a9e8b. Added a cancellation-aware FIFO submission gate shared by sendEmail and sendRawMessage and held through final reply or abort cleanup. Added a scripted integration test that withholds the first final reply while the server keeps reading and verifies a concurrent MAIL FROM cannot enter until the first transaction completes.

Sources/SwiftMail/SMTP/SMTPSendError.swift
jdudley: Fixed in d9a9e8b. Restored the payload-free Reason.timedOut case and moved the stage to the source-compatible optional SMTPSendError.timeoutStage property, including a defaulted public initializer parameter. Added compatibility and diagnostic tests.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
jdudley: Fixed in d9a9e8b. Content upload now records dispatch only after writeAndFlush is invoked; cancellation before the first buffer is classified notAccepted and retryable, while post-dispatch failures remain ambiguous. Added focused regression coverage for the pre-dispatch case.

Sources/SwiftMail/SMTP/SMTPSubmissionTimeouts.swift
jdudley: Fixed in d9a9e8b. Updated SMTPSubmissionTimeouts and GettingStartedWithSMTP to state that contentUpload is a per-64 KiB-buffer maximum reset after each successful flush; the final response retains a separate budget.

Sources/SwiftMail/SMTP/SMTPServer+Send.swift
```diff
@@ -43,6 +45,9 @@ extension SMTPServer {
      */
     @discardableResult
     public func sendEmail(_ email: Email) async throws -> SMTPSendResult {
+        try await submissionGate.acquire()
```
odrobnik: **HIGH** — Only `sendEmail` and `sendRawMessage` acquire this gate; public `login`, `authenticateXOAUTH2`, `fetchCapabilities`, `reset`, and `disconnect` still write to the same channel without it. During a backpressured multi-buffer DATA upload, a concurrent `login` can therefore enqueue `AUTH PLAIN <base64 credentials>` between two content buffers. The SMTP peer treats that line as message data, the original send can consume the eventual DATA `250` and report success, and the recipient receives a corrupted message containing reusable credentials. Serialize every public channel-writing/lifecycle operation against the complete submission transaction (with an internal ungated cleanup path where necessary), not just other send calls.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
```diff
@@ -0,0 +1,195 @@
...
+        ) {
+            if resolution.claim() {
+                completionPromise.fail(SMTPSubmissionTimeoutError(stage: timeoutStage))
+            }
+        }
+
+        let writeFuture: EventLoopFuture<Void> = channel.writeAndFlush(buffer)
+        // Record only after the buffer has actually been handed to NIO. A
+        // cancellation caught before this call remains provably pre-content.
+        contentDispatchState?.markContentDispatched()
```
odrobnik: **MEDIUM** — This flag becomes true after the first prefix buffer, even though `SendContentCommand` places the only unstuffed `<CRLF>.<CRLF>` terminator in the final buffer. If a large upload is cancelled or its transport fails between early buffers, that final buffer was never dispatched, so the server cannot have completed or accepted the message; nevertheless `classifyingContentFailure` returns `.ambiguous` / `.unsafeToRetry`. A durable outbox can then hold a provably unaccepted email for manual review instead of safely retrying it. Track whether the final terminator-bearing buffer was dispatched, rather than whether any content prefix was dispatched.

odrobnik (changes requested):
Requesting changes. The prior findings are partly addressed, but two safety defects remain: the transaction gate excludes other public command/lifecycle APIs, so credential-bearing AUTH (as well as EHLO, RSET, or QUIT) can still splice into a chunked DATA body; and the dispatch marker treats a failure after any prefix buffer as ambiguous even when the final SMTP terminator was never handed to NIO, producing a false unsafe-to-retry result. The latter conflicts with the end-of-data acceptance boundary in [RFC 5321 §3.3](https://www.rfc-editor.org/rfc/rfc5321.html#section-3.3). This was a read-only review; builds and tests were not run as requested.

- **HIGH** `Sources/SwiftMail/SMTP/SMTPServer+Send.swift:48`
- **MEDIUM** `Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift:173`

Sources/SwiftMail/SMTP/SMTPServer+Send.swift
jdudley: Fixed in 48d61e0. The send-only gate is now a cancellation-aware FIFO SMTPOperationGate covering connect, disconnect, login, XOAUTH2, capability refresh, reset, and complete send transactions. Nested workflows carry one unforgeable permit across command writes, pipeline mutation, final-reply waits, and cleanup, so no public channel operation can splice into DATA.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
jdudley: Fixed in 48d61e0. End-of-data is now marked only when the final buffer containing the completing CRLF is handed to NIO. Prefix-buffer timeout or cancellation remains proven not accepted and retryable; failures after final handoff remain ambiguous. Added partial-upload cancellation and 64 KiB boundary coverage.

jdudley:
Addressed both remaining findings in 48d61e0.

- Replaced the send-only gate with a cancellation-aware FIFO operation gate covering every public SMTP channel mutation, including connect/disconnect, authentication, EHLO, RSET, and complete mail submission. Nested workflows reuse one unforgeable permit across network awaits.
- End-of-data is now marked only when the final buffer containing the completing CRLF is handed to NIO. Failures before that boundary remain not accepted and retryable; failures afterward remain ambiguous and unsafe to retry.

I also replaced timing-based serialization assertions with deterministic gate-queue observation and added partial-upload cancellation and 64 KiB terminator-boundary coverage.

Validation:
- swift test: 449 tests across 57 suites passed
- swift build: passed
- swiftlint lint --strict --no-cache: 0 violations
- six serialization scenarios: 20/20 stress iterations passed

jdudley:
Follow-up: the production remediation remains the implementation in 48d61e0 (complete operation serialization plus end-of-data outcome classification).\n\nThe subsequent commits only stabilized the adversarial chunked-DATA test on Linux CI:\n\n- eac6238 makes the fake server’s DATA terminator scan incremental.\n- a73535a replaces the pathological 4 KiB receive window with a still-backpressured 256 KiB window and restores the original 5-second upload timeout.\n\nNo production SMTP behavior changed in those CI follow-ups. The fresh matrix is green on Linux, macOS, iOS, Android, and lint (Windows is intentionally skipped), all review threads are resolved, and the PR is mergeable.\n\nReady for re-review when convenient.

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
```diff
@@ -0,0 +1,228 @@
...
+                completionPromise.fail(SMTPSubmissionTimeoutError(stage: context.timeoutStage))
+            }
+        }
+
+        let writeFuture: EventLoopFuture<Void> = context.channel.writeAndFlush(buffer)
+        // Mark end-of-data only when the final buffer, including the completing
+        // CRLF, is handed to NIO. A prefix buffer may contain part of the
+        // terminator, but a failure before the final handoff is still provably
+        // pre-acceptance and remains safe to retry.
+        if dispatchesEndOfData {
```
odrobnik, line 222: **MEDIUM** — Large message upload over a lossy link. Prefix buffers 1–N succeed; the server then closes the connection (timeout or TCP RST). The final buffer's writeAndFlush returns an already-failed future because the channel is already closed. markEndOfDataDispatched() at line 204–205 sets dispatchedEndOfData = true unconditionally. classifyingContentFailure sees endOfDataWasDispatched: true and returns .ambiguous / .unsafeToRetry. A durable outbox holds the message for manual review even though the server provably never received the CRLF.CRLF terminator and retrying is safe.

odrobnik (changes requested):
The operation-gate serialization, chunked upload with per-buffer timeouts, and end-of-data outcome classification are coherent and address all prior findings. One correctness defect remains in the dispatch-marker placement.

**`markEndOfDataDispatched()` is set before `writeAndFlush` completes**

In `flushSubmissionBuffer`, `markEndOfDataDispatched()` is called on line 204–205, immediately after `context.channel.writeAndFlush(buffer)` is called but before `whenComplete` is registered and before the actor suspends to await the result. If `writeAndFlush` returns an already-failed future (channel closed between the last successful prefix-buffer write and the final buffer write), `dispatchedEndOfData` is unconditionally set to `true`. The failure then flows through `classifyingContentFailure(error, endOfDataWasDispatched: true)`, which returns `.ambiguous` / `.unsafeToRetry` — even though the server never received the completing `\r\n` and cannot have accepted the message.

The correct fix is to move the mark inside `whenComplete`, conditional on `.success`:

```swift
writeFuture.whenComplete { result in
    if resolution.claim() {
        timeoutTask.cancel()
        if dispatchesEndOfData, case .success = result {
            context.contentDispatchState?.markEndOfDataDispatched()
        }
        completionPromise.completeWith(result)
    }
}
```

This preserves the invariant stated in the PR description and in RFC 5321 §3.3: the server's acceptance boundary is crossed only when the terminator bytes are confirmed to have left the client's NIO stack. A connection drop before that confirmation remains `.notAccepted` / `.retryable`.

This was a read-only review; builds and tests were not run as requested.

- **MEDIUM** `Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift:204`

Sources/SwiftMail/SMTP/SMTPServer+SubmissionCommand.swift
jdudley: Fixed in 8404fd1. The final terminator write is now tracked as in flight before writeAndFlush, then recorded as succeeded or confirmed failed only when its completion callback wins the same timeout/cancellation resolution race. This makes an already-failed final write not-accepted/retryable as requested, while preserving an ambiguous/unsafe outcome if timeout or cancellation wins while that write is still unresolved. Added deterministic coverage for both the confirmed-failure and unresolved-write boundaries. Validation: swift build; swift test (451 tests); swift test --filter SMTPTests (110 tests); swiftlint lint --strict --no-cache (0 violations).

odrobnik:
@jdudley doing another review just now. Wanted to point out that your comments have a `\n` problem of double encoding or something. your newlines show as `\n` instead of being actual newlines.

odrobnik (approved):
Approved with no findings. The operation gate serializes every public SMTP channel mutation across the full submission and cleanup lifecycle; bounded DATA writes reset the upload budget per successful flush; the final-response budget starts only after the final flush; timeout stages remain source-compatible; and the final terminator state now distinguishes confirmed write failure from an unresolved timeout/cancellation race without creating an unsafe retry classification. Relevant implementation, call sites, handlers, public API, documentation, and regression coverage were inspected. Read-only review; builds and tests were not run as requested.

👍 No findings.

jdudley:
Thanks for pointing out the newline issue. I will remedy that going forward.

PR #210, opened 2026-08-10
Fix release test access to IMAP buffer limits

tabmail-kmyi:
## What

Keep the existing internal, read-only response-buffer test probes available in release builds as well as debug builds.

## Why

PR #179 added unconditional coverage for response-buffer propagation, but wrapped the two probes used by that suite in `#if DEBUG`. SwiftPM compiles the library without `DEBUG` under `swift test -c release`, while still compiling the test target, so the release test build fails before any tests can run.

This is a prerequisite for #208 so that its release-configuration validation can be evaluated on a green upstream baseline instead of being described as a pre-existing failure. Please land this small fix first; #208 can then incorporate the updated `main`.

## Impact

No public API or runtime behavior changes. The affected properties remain internal, read-only test accessors. This also keeps the buffer-limit suite active in release builds rather than skipping it.

## Checks

- `swift test` — 435 tests passed
- `swift test -c release` — 435 tests passed
- `swiftlint lint --strict --quiet`
- `git diff --check`
- independent audit — clean

Signed-off-by: Kwang Moo Yi <kmyi@tabmail.ai>

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -205,11 +205,9 @@ final class IMAPConnection {
         certificateVerificationPolicy
     }
 
-    #if DEBUG
     var responseBufferLimitForTesting: Int {
         responseBufferLimit
     }
-    #endif
 
     var namespacesSnapshot: NamespaceResponse? {
         namespaces
--- a/Sources/SwiftMail/IMAP/IMAPServer.swift
+++ b/Sources/SwiftMail/IMAP/IMAPServer.swift
@@ -288,12 +288,10 @@ public actor IMAPServer {
         return .plainText
     }
 
-    #if DEBUG
     /// Test-only access to the response buffer limit configured on the primary connection.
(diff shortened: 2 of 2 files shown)


tabmail-kmyi:
All checks are green. I also validated this exact four-line fix beneath the focused #208 review candidate: `swift test -c release` passed all 462 tests across 59 suites. I have held the #208 branch update so this prerequisite can land first, keeping the larger PR on a green release-test baseline.

odrobnik (approved):
No findings. The change correctly removes DEBUG-only compilation around two internal, read-only test probes so the unconditional `@testable` response-buffer tests can compile in release configuration. The probes have no side effects, remain outside the public API, and are referenced only by `IMAPResponseBufferLimitTests`; the primary, IDLE, and named connection paths and production parser setup continue to use the same stored buffer limit. Reviewed statically as requested; no builds or tests were run.

👍 No findings.

tabmail-kmyi:
Hi, can you merge this one? I'm planning to update #208 and it seems to require this PR first. THanks!

tabmail-kmyi:
Thanks for the quick action!

PR #214, opened 2026-08-15
Expose supportsMove, mirroring supportsUIDPlus

ZixTro:
`move(messages:to:)` falls back to `COPY` + `STORE \Deleted` + `EXPUNGE` when the server does not
advertise the MOVE extension. That is a reasonable default, but a caller cannot detect it: the
fallback is chosen internally and `capabilities` is not public.

This matters for clients with a no-delete policy. A bare `EXPUNGE` removes every message in the
mailbox already flagged `\Deleted`, including messages flagged by another client, so "archive this
one message" can silently delete unrelated mail on a server without MOVE.

This adds the same one-line accessor that already exists for UIDPLUS:

```swift
public var supportsMove: Bool { capabilities.contains(.move) }
```

No behaviour change; it only lets a caller decide whether to proceed or refuse.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPServer.swift
+++ b/Sources/SwiftMail/IMAP/IMAPServer.swift
@@ -90,6 +90,15 @@ public actor IMAPServer {
         capabilities.contains(.uidPlus)
     }
 
+    /// Whether the primary connection advertised MOVE (RFC 6851).
+    ///
+    /// Exposed so callers can tell an atomic `MOVE` from the `COPY` + `STORE \Deleted` + `EXPUNGE`
+    /// fallback that `move(messages:to:)` performs when the extension is missing. A client with a
+    /// no-delete policy needs to know which one it is about to get, and cannot currently ask.
+    public var supportsMove: Bool {
+        capabilities.contains(.move)
+    }
+
     var certificatePolicyForTesting: MailCertificateVerificationPolicy {
         primaryConnection.certificateVerificationPolicyForTesting
     }


Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -90,6 +90,15 @@ public actor IMAPServer {
         capabilities.contains(.uidPlus)
     }
 
+    /// Whether the primary connection advertised MOVE (RFC 6851).
+    ///
+    /// Exposed so callers can tell an atomic `MOVE` from the `COPY` + `STORE \Deleted` + `EXPUNGE`
+    /// fallback that `move(messages:to:)` performs when the extension is missing. A client with a
+    /// no-delete policy needs to know which one it is about to get, and cannot currently ask.
+    public var supportsMove: Bool {
```
odrobnik, line 98: **HIGH** — On a server advertising MOVE but not UIDPLUS, supportsMove returns true, but move(messages: UIDSet(...), to:) rejects the UID MOVE path because IMAPServer+Manipulation.swift:30 additionally requires UIDPLUS. It then performs UID COPY, UID STORE \\Deleted, and a mailbox-wide EXPUNGE at lines 153-157. If another client has an unrelated message already marked \\Deleted, a no-delete caller that proceeded because this accessor promised to distinguish the atomic path permanently deletes that unrelated message. Enforce the no-fallback decision inside move (and refresh/check capabilities there), rather than relying on this non-atomic raw Boolean.

Sources/SwiftMail/IMAP/IMAPServer.swift
```diff
@@ -90,6 +90,15 @@ public actor IMAPServer {
...
+        capabilities.contains(.move)
```
odrobnik, line 99: **MEDIUM** — IMAP capability tokens are case-insensitive, but the pinned NIOIMAPCore Capability preserves wire spelling and synthesizes case-sensitive Hashable equality. A conforming server that advertises `Move` or `move` produces a capability unequal to `.move`, so supportsMove reports false and a client refuses an atomic move that the server supports. Match the bare capability name case-insensitively and use the same predicate for dispatch.

odrobnik (changes requested):
Request changes: the new accessor does not reliably answer the no-delete safety question promised by its documentation. UID moves can still take the destructive fallback on a MOVE-only server, capability matching rejects valid mixed-case advertisements, and named connections remain unable to inspect their own MOVE capability. The safety policy needs an operation-level no-fallback guard, with case-insensitive capability handling and parity across both public connection surfaces.

- **HIGH** `Sources/SwiftMail/IMAP/IMAPServer.swift:98`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer.swift:99`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPNamedConnection.swift` — IMAPNamedConnection exposes the same public move(messages:to:) fallback and already mirrors supportsUIDPlus, but this change leaves it without supportsMove. A no-delete client operating through a named connection still has no way to detect that MOVE is absent; invoking named.move on a server without MOVE can therefore reach COPY + STORE \\Deleted + mailbox-wide EXPUNGE and delete unrelated mail. The primary connection's property is not authoritative for a separately authenticated or reconnected named connection.

Sources/SwiftMail/IMAP/IMAPServer.swift
odrobnik: Added the operation-level fallback policy on both move surfaces. With fallback disabled, move refreshes authentication state, accepts MOVE without UIDPLUS, and throws before COPY/STORE/EXPUNGE when MOVE is unavailable.

Sources/SwiftMail/IMAP/IMAPServer.swift
odrobnik: MOVE and UIDPLUS capability names are now matched as case-insensitive bare tokens, and the same MOVE predicate drives both supportsMove accessors and command dispatch.

odrobnik:
The superseding branch now includes the complete MOVE safety API on both `IMAPServer` and `IMAPNamedConnection`: case-insensitive `supportsMove`, plus `move(..., fallback: .disabled)` as an operation-level no-fallback guard that refreshes authentication state and permits UID MOVE without UIDPLUS. The original two-argument overloads retain their existing fallback behavior.

odrobnik:
MissionControl opened #225 from `agent/pr-214-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #225 merges. Thanks for the original contribution!

PR #215, opened 2026-08-20
Encode header values and display names that cannot be written literally

tabmail-kmyi:
## Problem

`String.rfc2047EncodedHeader()` decides whether to encode with:

```swift
guard self.contains(where: { !$0.isASCII }) else { return self }
```

That predicate is inverted with respect to the values that can actually break a header. **CR and LF are ASCII**, so a value carrying a raw CRLF is returned unchanged, while ordinary non-ASCII text — which cannot end a field — is safely wrapped in an encoded-word. The encoder passes through exactly the input that is dangerous and encodes the one that is not.

`Email+Content.writeHeaders` then emits `"Subject: \(self.subject.rfc2047EncodedHeader())\r\n"`, so a subject containing `\r\n` ends the `Subject:` field and everything after it is parsed as additional header fields:

```
Subject: Hi
Bcc: attacker@example.com
```

The predicate also iterates `Character` values. `CR-LF` is a *single* extended grapheme cluster, so `Character.isASCII` answers `true` for a whole `\r\n` pair — a `Character`-based test cannot see it even in principle.

Three further defects of the same family:

1. **Encoded-words can exceed the RFC 2047 §2 75-octet ceiling.** The word splitter flushes only *between* `Character`s. A grapheme cluster wider than the 45-byte budget has no interior boundary to flush at, so it lands alone in an oversized word. `"a"` followed by 23 combining marks is 47 UTF-8 bytes and yields a `12 + 4*ceil(47/3)` = **76-octet** word. Ordinary orthography cannot reach this; a long combining or ZWJ sequence in a user-supplied subject can.

2. **`EmailAddress.headerString()` carried the same inverted trigger**, and separately built its quoted-string by interpolation. An embedded `"` in a display name closes the quoted-string early and lets the remainder be read as further address syntax; a `\` is read as an RFC 5322 §3.2.4 quoted-pair, so the literal text is lost.

3. **`EmailAddress.description` is a second copy of the same formatter and was not covered by `headerString()` at all.** `Message.init(email:)` stores from/to/cc/bcc through `description`, and `EMLSerializer.writeHeaders` writes those strings into the header block verbatim — so a sender named `Bob\r\nBcc: attacker@example.com` produced a real `Bcc:` field in `emlData()`, reaching the same injection by a different path.

## What changed

**The encode trigger is decided per Unicode scalar** and covers everything RFC 5322 §2.2 excludes from a field body: below SP, plus DEL and above. HTAB is deliberately excluded — it is WSP, which RFC 5322 §3.2.5 permits literally in an unstructured field body, so ordinary values keep their exact output. Encoding rather than rejecting keeps the library non-throwing and loses nothing: an encoded-word is the standard carrier for octets a header cannot hold literally.

**The word splitter still walks `Character`s** — that is the correct unit whenever a cluster fits in a word at all, and it is what keeps ordinary text byte-identical — and descends to Unicode scalars *only* inside a cluster whose own UTF-8 length exceeds the 45-byte budget. A scalar is at most 4 bytes, so a chunk built from scalars never exceeds 45 bytes and the widest word it can produce is `12 + 4*ceil(45/3)` = 72 octets. Every word remains independently valid UTF-8, which is the property the existing `!chunk.isEmpty` guard exists to protect.

**A display name is also encoded when it holds a `"` or a `\`.** RFC 2047 §5 allows an encoded-word to replace a word inside a phrase, and its output is bare printable ASCII, so encoding both preserves the name and removes the escape. Encoded-words must not appear inside a quoted-string, so an encoded name is emitted bare, never quoted.

**`description` now returns `headerString()`**, so there is one formatter and every caller that formats an address gets a value a header field can hold.

**`init(_:)` decodes a *bare* display name**, which keeps the `LosslessStringConvertible` round trip intact for a name that now encodes, keeps `Email` → `Message` → `Email` lossless, and makes a `From:` line read off the wire yield the decoded name — the same treatment the IMAP `ENVELOPE` path already gives a `personName`. A display name inside a *quoted-string* is taken literally, because RFC 2047 §5 forbids reading an encoded-word there. That distinction is load-bearing rather than pedantic: `=?UTF-8?B?Qg0KQmNjOiBhQGIuY29t?=` decodes to a CRLF-bearing string, so decoding a quoted literal would manufacture the very injection this PR removes. A non-ASCII name never reaches the quoted branch, since `headerString()` emits it bare.

## Compatibility

**No API change.** No signature, access level or type is altered, so this is source-compatible including for code that stores these members as function values.

There is one deliberate **behaviour** change to a public member, stated plainly rather than buried:

> `EmailAddress.description` now returns an encoded-word for a display name holding non-ASCII text, a control character, a `"` or a `\`, where it previously returned that name raw or wrapped in a quoted-string the name itself could escape.

This is not a repurposing of the property. `description` has always produced RFC 5322 address syntax rather than free text — a name containing a comma already came back quoted — so encoding a name that syntax cannot carry literally is the same contract applied to the values that were escaping it. If you would rather `description` stay byte-for-byte as it was and only `headerString()` be hardened, say so and I will split it; the consequence is that `Message(email:)` → `emlData()` keeps the injection, since that path formats through `description`.

Every value that could already be emitted correctly is returned **byte-identically**, and that claim is pinned by a test that transcribes the exact words this encoder produced beforehand for two emoji sequences, rather than merely asserted.

## Tests

16 tests added, all asserted on final serialized bytes or on what a conforming reader recovers from them, never on the helper in isolation. Red-first was measured, not assumed:

| commit | added | red against the parent | non-vacuity / guard |
|---|---|---|---|
| Encode header values… | 9 | **7** | 2 |
| Format every address… | 7 | **5** | 2 |

The red cases include a sweep over every C0 control and DEL, the 76-octet word, the injected `Bcc:` line from both a subject and a display name, the truncated display names, the forged `From:` carrying two addresses, and the round trip losing the name.

The non-vacuity twins are the reason to trust the rest: `"a b\tc"`, `"Re: Q3 report"` and a value containing `"` and `\` are each returned byte-identical before and after, so the trigger cannot have been widened into ordinary text. A two-sided pin holds the decode rule from both ends — a quoted encoded-word-shaped literal is preserved *and* a bare one still decodes — so neither half can be disarmed without failing the other.

Both suites in `RFC2047EncodeTests.swift` also gain the `.serialized, .timeLimit(.minutes(1))` traits the rest of the test targets use, so a hang fails in a minute and says so instead of consuming the job timeout. Happy to drop that hunk if you would prefer it separately.

## Validation

- `swift test` — **511 tests in 64 suites passed** (up from 495 in 62 on `main`)
- `swift test -c release` — **511 tests in 64 suites passed**
- `swift build` — no warnings in `Sources/SwiftMail`; the two `Demos` `LogHandler` deprecations are pre-existing and untouched
- `swiftlint lint --strict` — 0 violations across the changed files
- Branched from `a2d4a94`, re-confirmed as `main` immediately before opening this PR

## Deliberately left out

Adjacent surfaces of the same class, not touched here so this PR stays one concern. Happy to follow up on any of them:

- **`Email.writeHeaders` interpolates caller-supplied `additionalHeaders` raw** — `content += "\(key): \(value)\r\n"`, with neither key nor value validated. This is the largest remaining hole, but the right remedy is a validation/rejection decision on arbitrary header names rather than encoding, so it needs an API call you should make.
- **A CR or LF in `EmailAddress.address`** — a validation concern rather than an encoding one; an encoded-word is not legal in an `addr-spec`.
- **`MessageID` components**, which reach `Message-Id:` by the same interpolation.
- **RFC 5322 §2.1.1's 998-octet line limit** for long all-ASCII values, which no folding currently enforces.

Boundaries are *not* on this list: every one is library-generated (`UUID().uuidString` / `generateBoundary()`), so no caller or remote sender can steer them.

The **MIME part-header** parameters (`filename`, `name`, and the content types beside them) are the same class of defect one level down, and are fixed in the sibling PR below rather than here.

## Note

This is one of a pair with #216, which fixes the MIME part-header parameter serializer and parser — filename and content-type interpolation, with an RFC 2231 policy. The two share no files and can be reviewed and merged in either order.

## Follow-up: preserve validated folding whitespace

Cross-component validation found one compatibility edge: callers may already supply RFC 2047 encoded-words separated by legal RFC 5322 folding whitespace. Treating every CR/LF scalar as unsafe re-encoded that already-valid ASCII value.

The encoder now preserves a fold only when it is a complete `CRLF` followed by SP/HTAB whose whitespace run leads to printable ASCII. Bare LF, lone CR, trailing folds, consecutive folds without intervening visible content, and a later malformed line break still take the encoded path. This keeps header-injection protection intact while making pre-folded encoded-words byte-stable.

The follow-up is intentionally narrow: 35 production additions and 36 test additions in the existing encoder/test files. A targeted correctness, robustness/security, and mutation-oriented test re-gate is clean; the focused RFC 2047 run passes 27 tests across 4 suites, strict SwiftLint reports 0 violations, and the combined fork pin passes 559 tests across 70 suites. Upstream CI for this updated head is running.

### DIFF
--- a/Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
+++ b/Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
@@ -8,6 +8,15 @@ import Foundation
 extension EmailAddress: LosslessStringConvertible {
     /**
      Initialize an email address from a string representation
+
+     A *bare* encoded-word display name is RFC 2047-decoded — that is how a
+     non-ASCII name written by ``description``, or read off the wire, arrives —
+     so the round trip yields the name the recipient actually sees, the same
+     treatment the IMAP `ENVELOPE` path gives a `personName`. A display name
+     inside a *quoted-string* is taken literally: RFC 2047 §5 forbids reading an
+     encoded-word there, so text that merely looks like `=?…?=` is returned
+     verbatim, not decoded.
+
      - Parameter description: The string representation of the email address
      */
     public init?(_ description: String) {
@@ -34,14 +43,21 @@ extension EmailAddress: LosslessStringConvertible {
 
                 // Check if we have a quoted name or a regular name
                 if nameRange1.location != NSNotFound {
-                    // Quoted name (with special characters)
+                    // Quoted name (with special characters). A quoted-string
+                    // always carries a LITERAL display name: RFC 2047 §5 forbids
+                    // reading an encoded-word inside a quoted-string, so a name
+                    // that merely *looks* like `=?UTF-8?B?…?=` is exactly that
+                    // text and must be handed back verbatim, never MIME-decoded.
+                    // A non-ASCII name never reaches this branch — `headerString`
+                    // emits it as a *bare* encoded-word, which the unquoted
+                    // branch below decodes.
                     let name = nsString.substring(with: nameRange1)
                     self.init(name: name, address: email)
                     return
                 } else if nameRange2.location != NSNotFound {
                     // Regular name
                     let name = nsString.substring(with: nameRange2).trimmingCharacters(in: .whitespaces)
-                    self.init(name: name, address: email)
+                    self.init(name: name.decodeMIMEHeader(), address: email)
                     return
                 } else {
                     // Just the email
@@ -56,20 +72,17 @@ extension EmailAddress: LosslessStringConvertible {
 
     /**
      Get the string representation of the email address
-     This uses the formatted representation which includes the name if available
+
+     This is the RFC 5322 address string, including the display name if there is
+     one — the same text ``headerString()`` writes into a header field, and the
+     text ``init(_:)`` reads back. It has always produced address syntax rather
+     than free text (a name with a comma comes back quoted), so a name that
+     syntax cannot carry literally is RFC 2047-encoded here too; see
+     ``headerString()`` for which names those are and why. Emitting a name raw
+     when it holds a CR or LF is what let a `Message` built from an `Email` grow
+     a header field its author never wrote.
      */
-    public var description: String {
-        if let name = name, !name.isEmpty {
-            // Use quotes if the name contains special characters
-            if name.contains(where: { !$0.isLetter && !$0.isNumber && !$0.isWhitespace }) {
-                return "\"\(name)\" <\(address)>"
-            } else {
-                return "\(name) <\(address)>"
-            }
-        } else {
-            return address
-        }
-    }
+    public var description: String { headerString() }
 
     /**
      RFC 5322 address string for use in a header field (`From`/`To`/`Cc`/…).
@@ -90,7 +103,9 @@ extension EmailAddress: LosslessStringConvertible {
      removes the escape. Encoded-words must not appear inside a quoted-string, so
      an encoded name is emitted bare (never quoted).
 
-     Use this — not ``description`` — when writing an address into a header.
+     ``description`` returns this, so every caller that formats an address gets a
+     value a header field can hold. ``init(_:)`` decodes the name again, so the
+     round trip still yields the original text.
      */
     func headerString() -> String {
         guard let name = name, !name.isEmpty else { return address }
--- a/Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
+++ b/Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
@@ -74,18 +74,30 @@ extension EmailAddress: LosslessStringConvertible {
     /**
      RFC 5322 address string for use in a header field (`From`/`To`/`Cc`/…).
 
-     Identical to ``description`` for ASCII display names, but a non-ASCII name
-     is RFC 2047-encoded. Encoded-words must not appear inside a quoted-string,
-     so an encoded name is emitted bare (never quoted). Use this — not
-     ``description`` — when writing an address into a header so that non-ASCII
-     names survive transport instead of being mojibake'd.
+     A display name that cannot be written literally is RFC 2047-encoded. That
+     covers two cases:
+
+     - The name is not a valid field body — it holds non-ASCII text or a control
+       character. A CR or LF here *ends the header field*, so an unencoded name
+       turns the rest of the value into new header lines.
+     - The name holds a `"` or a `\`, the two characters that would escape the
+       quoted-string it would otherwise be wrapped in. An embedded `"` closes the
+       string early and lets the remainder be read as further address syntax; a
+       `\` is read as a quoted-pair, so the literal text is lost.
+
+     An encoded-word may replace a word inside a phrase (RFC 2047 §5) and its
+     output is bare printable ASCII, so encoding both carries the name intact and
+     removes the escape. Encoded-words must not appear inside a quoted-string, so
+     an encoded name is emitted bare (never quoted).
+
+     Use this — not ``description`` — when writing an address into a header.
      */
     func headerString() -> String {
         guard let name = name, !name.isEmpty else { return address }
-        if name.contains(where: { !$0.isASCII }) {
-            return "\(name.rfc2047EncodedHeader()) <\(address)>"
+        if name.rfc2047RequiresEncodingAsDisplayName {
+            return "\(name.rfc2047EncodedWords()) <\(address)>"
         }
-        // Use quotes if the (ASCII) name contains special characters
+        // Use quotes if the (plain ASCII) name contains special characters
         if name.contains(where: { !$0.isLetter && !$0.isNumber && !$0.isWhitespace }) {
             return "\"\(name)\" <\(address)>"
         }
--- a/Sources/SwiftMail/Extensions/String+RFC2047Encode.swift
+++ b/Sources/SwiftMail/Extensions/String+RFC2047Encode.swift
@@ -40,6 +40,40 @@ extension String {
         return scalar.value < 0x20 || scalar.value >= 0x7F
     }
 
+    /// Whether the receiver contains a scalar that cannot be written literally
+    /// in an unstructured header field body. A CRLF pair is allowed only when
+    /// followed by whitespace: that is RFC 5322 folding whitespace, not a new
+    /// header field. This preserves already-folded encoded-words supplied by a
+    /// caller while bare or malformed line breaks still trigger encoding.
+    private var rfc2047ContainsUnsafeHeaderScalar: Bool {
+        let scalars = unicodeScalars
+        var index = scalars.startIndex
+
+        while index != scalars.endIndex {
+            let scalar = scalars[index]
+            if scalar == "\r" {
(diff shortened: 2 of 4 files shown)


Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
```diff
@@ -56,36 +72,47 @@ extension EmailAddress: LosslessStringConvertible {
...
-            if name.contains(where: { !$0.isLetter && !$0.isNumber && !$0.isWhitespace }) {
-                return "\"\(name)\" <\(address)>"
-            } else {
-                return "\(name) <\(address)>"
-            }
-        } else {
-            return address
-        }
-    }
+    public var description: String { headerString() }
```
odrobnik, line 85: **HIGH** — Create an Email whose sender name is `Bob\r\nBcc: attacker@example.com\r\nX-Ignore:` and address is `real@example.com`. `Message(email:).emlData()` is initially safe, but `Message(emlData:)` decodes the bare encoded-word back to raw CRLF in `MessageInfo.from`; calling `emlData()` on that parsed Message writes the value verbatim and produces a real `Bcc: attacker@example.com` field, with `X-Ignore` absorbing the real addr-spec. The new formatter therefore protects only the first serialization; parsed address fields must remain structured/wire-safe or be re-encoded before every serialization.

odrobnik, line 85: **HIGH** — For `EmailAddress(name: "홍<evil@example.com>", address: "real@example.com")`, the first EML safely contains an encoded phrase. `Message(emlData:)` then decodes the entire structured `From` value before mailbox parsing, yielding `홍<evil@example.com> <real@example.com>`; the unanchored `EmailAddress` regex selects the display name's first angle pair, so `Email(message:)` restores `evil@example.com` as the sender. A storage/parse round trip can thus replace the real envelope sender; address structure must be parsed before display-name decoding.

odrobnik, line 85: **MEDIUM** — A display name longer than the 45-byte word budget (for example 16 Hangul syllables) makes `description` contain multiple encoded-words separated by `CRLF SPACE`. Swift counts CRLF as one `Character`, while `NSRegularExpression` ranges are UTF-16; the existing `NSRange(location:length: description.count)` therefore excludes the final `>` and `EmailAddress(address.description)` returns nil. Direct `Message(email:) -> Email(message:)` then throws for a long sender, while long To/Cc/Bcc entries are silently dropped by `compactMap`, violating the new LosslessStringConvertible round-trip.

Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
```diff
@@ -34,14 +43,21 @@ extension EmailAddress: LosslessStringConvertible {
...
+                    // emits it as a *bare* encoded-word, which the unquoted
+                    // branch below decodes.
                     let name = nsString.substring(with: nameRange1)
                     self.init(name: name, address: email)
                     return
                 } else if nameRange2.location != NSNotFound {
                     // Regular name
                     let name = nsString.substring(with: nameRange2).trimmingCharacters(in: .whitespaces)
-                    self.init(name: name, address: email)
+                    self.init(name: name.decodeMIMEHeader(), address: email)
```
odrobnik, line 60: **MEDIUM** — `pre=?UTF-8?B?SGVsbG8=?=post <bob@example.com>` contains one ordinary RFC 5322 atom; the encoded-word-shaped substring is not a complete phrase word and must remain literal. The new `decodeMIMEHeader()` call regex-decodes the interior substring anyway, returning the display name `preHellopost` (and an analogous payload can manufacture CRLF), corrupting input that the initializer previously preserved. Decode only complete encoded-word tokens in valid phrase positions.

odrobnik (changes requested):
Request changes: the first serialization is hardened, but parsed address values are decoded back into raw control/structured text and later reused as wire syntax, enabling header reinjection and sender substitution. Long folded display names also fail or mutate across the promised round trips, and the new initializer decodes encoded-word-looking substrings that are not RFC 2047 phrase tokens.

One review with 5 findings — those on changed lines are detailed in inline comments:

- **HIGH** `Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift:85`
- **HIGH** `Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift:85`
- **MEDIUM** `Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift:85`
- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+RFC2047.swift` — After `Message(email:).emlData()` emits adjacent folded encoded-words for a long display name, header unfolding changes their separator to a space. `EMLParser.decodeRFC2047` decodes each regex match independently but retains that separator, although RFC 2047 requires whitespace between adjacent encoded-words to be ignored. A 16-syllable Hangul name therefore returns as 15 syllables, an inserted space, and the final syllable after `Message(emlData:) -> Email(message:)`.
- **MEDIUM** `Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift:60`

Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
odrobnik: Fixed by decoding only complete RFC 2047 encoded-word tokens in bare phrase positions. Encoded-word-shaped substrings embedded inside an atom now remain literal, including a payload that would decode to CRLF.

Sources/SwiftMail/Extensions/EmailAddress+StringConversion.swift
odrobnik: Fixed by parsing the final anchored angle-addr before decoding the display name, and by no longer RFC 2047-decoding whole address fields in EMLParser. The real addr-spec now survives the malicious display-name angle pair.

odrobnik: Fixed by keeping parsed address fields in structured wire form and re-hardening every From/To/Cc/Bcc value in EMLSerializer before emission. Added a regression covering parse followed by a second serialization of the CRLF-bearing sender name.

odrobnik: Fixed by replacing the mixed Character-count/UTF-16 regex range with native String-index parsing anchored at the final angle-addr. Added direct and Email/Message/EML round trips for a folded 16-syllable Hangul name across sender, To, Cc, and Bcc.

odrobnik:
MissionControl opened #226 from `agent/pr-215-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #226 merges. Thanks for the original contribution!

tabmail-kmyi:
Welcome! Got drowned by other issues and was gonna come
back to it later, but i guess that took longer than i hoped! Thanks for taking care of it and also the great software.

PR #216, opened 2026-08-20
Encode MIME parameters and anchor parameter parsing to the RFC 2045 grammar

tabmail-kmyi:
## Problem

Two halves of the same defect: the library writes MIME header parameters by string interpolation, and reads them back with a substring search. Both treat sender- or caller-chosen text as if it were grammar.

### Writing

Every place the library writes a part header interpolates the value directly:

```swift
contentType += "; name=\"\(filename)\""
dispValue   += "; filename=\"\(filename)\""
content     += "Content-Type: \(attachment.mimeType)\r\n"
content     += "Content-ID: <\(contentID)>\r\n"
```

`Attachment` and `MessagePart` store what the caller gave them verbatim and nothing validates in between, so:

- a `"` in a filename closes the quoted-string and the remainder is read as further parameters — a filename of `a"; boundary="X.pdf` rewrites the part's parameter list;
- a CR or LF in a filename, content type, disposition or Content-ID ends the field, and everything after it is read as new header fields (`Bcc:` of the attacker's choosing);
- a non-ASCII filename goes out as raw UTF-8 inside a quoted-string, which RFC 5322 §2.2 does not allow a header field body to carry.

There are **six** filename sites (two in `EMLSerializer.serializePartHeaders`, two in `Email.writeMultipartMixed`, two in `Email.writeHTMLWithInlineAttachments`) and **four** content-type sites, each hand-rolled.

### Reading

`EMLParser.extractHeaderParam` located a parameter by searching for `<name>=` as a plain substring, and `cleanContentType` split on every `;`. Neither knows the RFC 2045 §5.1 grammar, so structure placed inside a legal quoted value was obeyed:

```
multipart/mixed; name="x boundary=forged"; boundary=real
```

reported the boundary as `forged"` rather than `real` — a sender choosing how a multipart body is split. And `application/pdf; name="a;b.pdf"` was stored as the corrupt `application/pdf; b.pdf"`, because a `;` is perfectly legal between the quotes.

These halves are one PR because they are one round trip: the writer must emit an RFC 2231 extended parameter for values a quoted-string cannot carry, and the parser must read that spelling back, or the library cannot parse its own output.

## What changed

**Parser first** (commit 1), so each commit is correct on its own and the series bisects.

`parameterSegments(of:)` splits a field body into its top-level `;`-separated segments while tracking quoted-string state, and an attribute now matches only when it is the *whole* attribute of a segment. Both the splitter and the quoted-string value reader honour an RFC 2045 quoted-pair — inside the quotes a `\` makes the next character literal — so the two agree on where a value ends. If they did not, a parameter could be found in a segment whose value the reader then runs past: `attachment; name="a\"; name*=UTF-8''evil.pdf"` is *one* parameter whose value is the whole literal, and a splitter toggling on every `"` would re-open the split early and expose the interior `name*`.

`extractFilename` gains the RFC 2231 §4 extended spelling of `name` alongside the `filename*` it already decoded, and the resolution order is now written once as data, ranked by **attribute first and spelling second**: `filename*`, `filename`, `name*`, `name`. RFC 2183 §2.3 ranks the attributes — `filename` names the file, `name` is the deprecated Content-Type spelling. RFC 2231 §4 ranks only the two spellings of *one* attribute; it does not rank `name*` over `filename`. The order is resolved across both headers at once, because a per-header order cannot express a ranking between attributes living in different headers.

**Writer** (commit 2). All six filename sites go through one shared `MIMEHeaderEncoding.parameter(name:value:)`. A value that can be written literally inside a quoted-string is emitted as `name="value"`, byte for byte as before; anything else is emitted as a single RFC 2231 §4 extended parameter, `name*=UTF-8''<percent-encoded>`. **Exactly one of the two forms, selected by the value, never both**, so a receiver has nothing to choose between. The extended form is deliberately a single segment — RFC 2231 §3 continuations (`name*0*=`) are legal but far less widely implemented, and the value has to survive one header line.

The content-type, Content-ID, disposition and transfer-encoding values are different in kind: they legitimately carry their own parameter syntax (`text/calendar; method=REQUEST`, `charset=…`), so they cannot be wrapped in an extended parameter without destroying it. They go through `MIMEHeaderEncoding.fieldBody`, which removes the scalars RFC 5322 §2.2 excludes from a field body and returns an already-valid value unchanged.

> **A design question for you.** `fieldBody` *strips* the offending scalars. The stricter alternative is to reject the value, but that would make `constructContent()` throwing, so this PR takes the non-breaking option. If you would rather it throw, that is a one-commit change on top and I am happy to make it — it is your API compatibility call, not mine.

Membership in both predicates is decided on the Unicode **scalar**, never on `Character`: CR-LF is a single extended grapheme cluster, so `Character.isASCII` answers `true` for a whole `\r\n` pair and cannot see it. The two predicates disagree over HTAB deliberately, and the reason is documented at the point of disagreement — a parameter value has an encoded form to fall back to and round-trips HTAB exactly as `%09`, whereas a field body that must keep its own parameter syntax has none and keeps it as the legal WSP it is.

## Compatibility

**No public API change.** `MIMEHeaderEncoding` is internal; `EMLParser.extractFilename(from:)` becomes variadic to take both headers, but it too is internal. No public signature, access level or type is altered, so this is source-compatible including for function values.

Output is **byte-identical** for every value that could already be emitted correctly — pinned by four non-vacuity tests rather than asserted.

One deliberate **receive-side behaviour** change, stated plainly:

> A part whose Content-Type and Content-Disposition carry **different** filenames now resolves to the disposition's `filename`, where it previously resolved to the Content-Type's `name`.

That is the same RFC 2183 §2.3 ranking rather than a second rule, and it can only decide a part whose two headers disagree. Where they agree — the ordinary case, and what this library's own serializer emits — there is no difference.

## Tests

33 tests added, asserted on the final serialized bytes and on what `Message(emlData:)` recovers from them, never on the encoder in isolation. Red-first was measured, not assumed:

| commit | added | red against the parent | green guard / twin |
|---|---|---|---|
| Anchor MIME parameter parsing… | 13 | **8** | 5 |
| Encode MIME parameters… | 20 | **16** | 4 |

The parser reds are the forged boundary, the corrupt content type, the two escaped-quote spoofs, the value truncated at its quoted-pair, the cross-header ranking, and the two `name*` cases (red because the spelling had no support at all, not because it was mis-handled). The writer reds are the forged-parameter filename coming back as `"a"`, the Korean filename putting 8-bit octets in a header, and a CRLF in a filename, content type, calendar content type, disposition or Content-ID producing a `Bcc:` line.

The green ones are the reason to trust the rest. On the parser side they hold the attribute ranking from *both* directions, so the two precedence rules cannot be collapsed into one, and one of them is precisely why the extended spelling can be added safely: an attribute inside a quoted value must not be read as a parameter — something the base satisfied only by accident, because it could not decode `name*` anywhere. On the writer side an ordinary filename, an ordinary part filename, `text/plain; charset=windows-1252`, and a `;` inside a quoted filename all stay byte-identical.

Coverage deliberately includes sites reachable only through the library's own API and not through any one caller's path — the `writeHTMLWithInlineAttachments` inline-attachment `name`/`filename` pair and both Content-ID sites — so the whole surface is pinned rather than the parts one application happens to exercise.

## Validation

- `swift test` — **528 tests in 65 suites passed** (up from 495 in 62 on `main`)
- `swift test -c release` — **528 tests in 65 suites passed**
- `swift build` — no warnings in `Sources/SwiftMail`; the two `Demos` `LogHandler` deprecations are pre-existing and untouched
- `swiftlint lint --strict` — 0 violations across the changed files
- Branched from `a2d4a94`, re-confirmed as `main` immediately before opening this PR

## Deliberately left out

Kept out so this PR stays one concern; happy to follow up:

- **`Email.writeHeaders` interpolates caller-supplied `additionalHeaders` raw** — `content += "\(key): \(value)\r\n"`, neither key nor value validated. The remedy is a validation decision on arbitrary header names rather than encoding, so it needs an API call from you.
- **RFC 2231 §3 continuations** for parameter values too long for one line — the writer emits a single segment and the reader accepts a single segment. Worth doing if you want long non-ASCII filenames folded rather than carried on one line.
- **`MessageID` components** and a CR/LF in `EmailAddress.address`, both of which reach the top-level header block by the same interpolation.

Boundaries are *not* on this list: every one is library-generated (`UUID().uuidString` / `generateBoundary()`), so no caller or remote sender can steer them.

## Note

This is one of a pair with #215, which fixes the RFC 2047 encoder and the address display-name formatter for the same class of defect in the *top-level* header block. The two share no files and can be reviewed and merged in either order.

### DIFF
--- a/Sources/SwiftMail/MIME/EMLParser+Parameters.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+Parameters.swift
@@ -13,14 +13,66 @@ extension EMLParser {
         return parts.first.map { String($0).trimmingCharacters(in: .whitespaces) } ?? contentType
     }
 
+    /// Split a header field body into its top-level `;`-separated segments.
+    ///
+    /// A `;` between the quotes of a parameter value is part of that value —
+    /// `name="a;b.pdf"` is one parameter, not two — so splitting has to know
+    /// where the quotes are, or a value the sender chooses decides where the
+    /// parameters are.
+    ///
+    /// An RFC 2045 quoted-pair inside the quotes — a `\` before any character —
+    /// is honored: the escaped character stands for itself, so a `\"` does not
+    /// close the value. The quoted-string reader in
+    /// ``extractHeaderParam(from:named:)`` applies the identical rule, so the
+    /// two agree on where a value ends; if they did not, a parameter could be
+    /// found in a segment whose value the reader then runs past.
+    static func parameterSegments(of header: String) -> [Substring] {
+        var segments: [Substring] = []
+        var start = header.startIndex
+        var index = header.startIndex
+        var inQuotes = false
+
+        while index < header.endIndex {
+            let character = header[index]
+            if inQuotes && character == "\\" {
+                // Quoted-pair: the next character is literal, so it can neither
+                // close the quote nor separate parameters. Skip both here, as
+                // the value reader unescapes them, or the two disagree on where
+                // the quoted value ends.
+                index = header.index(after: index)
+                if index < header.endIndex {
+                    index = header.index(after: index)
+                }
+                continue
+            }
+            if character == "\"" {
+                inQuotes.toggle()
+            } else if character == ";" && !inQuotes {
+                if start < index {
+                    segments.append(header[start..<index])
+                }
+                start = header.index(after: index)
+            }
+            index = header.index(after: index)
+        }
+
+        if start < header.endIndex {
+            segments.append(header[start...])
+        }
+        return segments
+    }
+
     /// Clean a Content-Type value for storage in MessagePart.
     /// Preserves charset and other relevant params, strips name/filename/boundary.
     static func cleanContentType(_ contentType: String) -> String {
-        let components = contentType.split(separator: ";")
+        let components = parameterSegments(of: contentType)
         guard let mimeType = components.first else { return contentType }
 
         var result = String(mimeType).trimmingCharacters(in: .whitespaces)
-        let skipParams: Set<String> = ["name", "filename", "boundary"]
+        // The RFC 2231 extended spellings are skipped alongside the plain ones,
+        // so a name that had to be written as `name*=` is not left in the stored
+        // content type for a later re-serialization to append a second copy of.
+        let skipParams: Set<String> = ["name", "filename", "boundary", "name*", "filename*"]
 
         for component in components.dropFirst() {
             let trimmed = String(component).trimmingCharacters(in: .whitespaces)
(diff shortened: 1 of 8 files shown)


Sources/SwiftMail/MIME/EMLParser+Parameters.swift
```diff
@@ -13,14 +13,66 @@ extension EMLParser {
...
+    /// two agree on where a value ends; if they did not, a parameter could be
+    /// found in a segment whose value the reader then runs past.
+    static func parameterSegments(of header: String) -> [Substring] {
+        var segments: [Substring] = []
+        var start = header.startIndex
+        var index = header.startIndex
+        var inQuotes = false
+
+        while index < header.endIndex {
+            let character = header[index]
```
odrobnik, line 36: **HIGH** — RFC 6532 permits UTF-8 qtext, but iterating Swift `Character`s makes an ASCII quote followed by U+0301 one grapheme that is not equal to `"`. For `multipart/mixed; x="́a; boundary=evil"; boundary=real`, the inner semicolon is split and the real separator is hidden, so `extractBoundary` returns `evil"; boundary=real`; a remote sender can include matching fake delimiter lines and make SwiftMail construct an attacker-selected MIME part tree.

Sources/SwiftMail/MIME/EMLParser+Parameters.swift
```diff
@@ -13,14 +13,66 @@ extension EMLParser {
...
+                // close the quote nor separate parameters. Skip both here, as
+                // the value reader unescapes them, or the two disagree on where
+                // the quoted value ends.
+                index = header.index(after: index)
+                if index < header.endIndex {
+                    index = header.index(after: index)
+                }
+                continue
+            }
+            if character == "\"" {
```
odrobnik, line 48: **MEDIUM** — RFC 2045 allows RFC 822 comments in Content-Type and requires processors to ignore them. A valid `Content-Type: multipart/mixed (size 6"); boundary=outer` contains a quote only inside a comment, but this toggles `inQuotes` and suppresses the boundary separator, causing `Message(emlData:)` to return the multipart body as one opaque part instead of its text and attachments. A semicolon inside a comment can likewise be exposed as a forged parameter.

Sources/SwiftMail/MIME/EMLParser+Parameters.swift
```diff
@@ -40,46 +92,120 @@ extension EMLParser {
...
+        return nil
+    }
+
+    /// Extract an RFC 2231 extended parameter (`name*=charset'language'value`)
+    /// and percent-decode its value. A single segment only — continuations
+    /// (`name*0*=`) are not read.
+    static func extractExtendedHeaderParam(from header: String, named name: String) -> String? {
+        guard let raw = extractHeaderParam(from: header, named: name + "*") else { return nil }
+        // Strip the encoding prefix, e.g. "UTF-8''filename.txt".
+        guard let separator = raw.range(of: "''") else { return raw }
```
odrobnik, line 206: **MEDIUM** — For the valid `Content-Type: application/pdf; name="fallback.pdf"; name*=UTF-8'en'r%C3%A9sum%C3%A9.pdf`, searching only for two adjacent apostrophes returns the raw `UTF-8'en'...` token. Because this change newly ranks `name*` above `name`, callers receive a corrupt filename instead of the decoded `résumé.pdf` (or the usable fallback).

Sources/SwiftMail/MIME/MIMEHeaderEncoding.swift
```diff
@@ -0,0 +1,113 @@
...
+    /// single RFC 2231 §4 extended parameter, `name*=UTF-8''<percent-encoded>`.
+    ///
+    /// Exactly one of the two forms is emitted, selected by the value; the two
+    /// are never emitted side by side, so a receiver has nothing to choose
+    /// between. The extended form is deliberately a single segment: RFC 2231 §3
+    /// continuations (`name*0*=`, `name*1*=`) would be legal but are far less
+    /// widely implemented, and the value has to survive one header line.
+    static func parameter(name: String, value: String) -> String {
+        guard value.unicodeScalars.allSatisfy(isSafeInQuotedString) else {
+            return "\(name)*=UTF-8''\(percentEncoded(value))"
```
odrobnik, line 30: **MEDIUM** — A filename containing 110 Korean syllables expands to about 990 characters in the `%XX` value alone; after `Content-Disposition: attachment; filename*=UTF-8''` is added, the emitted physical line exceeds RFC 5322's 998-octet hard limit. With no folding, continuation, or rejection downstream, conforming SMTP infrastructure may reject or truncate the message.

Sources/SwiftMail/MIME/MIMEHeaderEncoding.swift
```diff
@@ -0,0 +1,113 @@
...
+    /// such as `text/calendar; method=REQUEST`, a Content-ID — cannot be wrapped
+    /// in an encoded-word or an extended parameter without destroying that
+    /// syntax, so the scalars a field body cannot hold are removed instead.
+    ///
+    /// A value that is already a valid field body is returned unchanged, so this
+    /// only ever alters a value that could not have been emitted correctly.
+    static func fieldBody(_ value: String) -> String {
+        guard value.unicodeScalars.contains(where: isForbiddenInFieldBody) else { return value }
+        var scalars = String.UnicodeScalarView()
+        for scalar in value.unicodeScalars where !isForbiddenInFieldBody(scalar) {
```
odrobnik, line 49: **MEDIUM** — Round-tripping an RFC 6532 part with `Content-ID: <café@example.com>` through `Message.emlData()` now emits `<caf@example.com>` because `é` is silently deleted, while the HTML body still references `cid:café@example.com`. The inline resource no longer resolves, and distinct valid IDs can collapse to the same serialized value.

odrobnik (changes requested):
Requesting changes for five correctness and security gaps in the new MIME grammar and encoding path: the parameter scanner can misread valid UTF-8 quoted strings and RFC 822 comments as syntax; RFC 2231 values with a language tag are not decoded; single-segment percent encoding can exceed the mandatory header-line limit; and field-body sanitization deletes valid RFC 6532 UTF-8 Content-ID data. References: [RFC 2045 §5.1](https://www.rfc-editor.org/rfc/rfc2045.html#section-5.1), [RFC 2231 §4](https://www.rfc-editor.org/rfc/rfc2231.html#section-4), and [RFC 6532 §3.2](https://www.rfc-editor.org/rfc/rfc6532.html#section-3.2). No builds or tests were run, per the review constraints.

One review with 5 findings, each detailed in an inline comment:

- **HIGH** `Sources/SwiftMail/MIME/EMLParser+Parameters.swift:36`
- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+Parameters.swift:48`
- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+Parameters.swift:206`
- **MEDIUM** `Sources/SwiftMail/MIME/MIMEHeaderEncoding.swift:30`
- **MEDIUM** `Sources/SwiftMail/MIME/MIMEHeaderEncoding.swift:49`

Sources/SwiftMail/MIME/EMLParser+Parameters.swift
odrobnik: Fixed the MIME parameter scanner and quoted-string reader to operate on Unicode scalars, so ASCII quotes adjacent to combining marks are recognized as grammar. Added regressions for the forged-boundary case and quoted values beginning with a combining mark.

odrobnik: Fixed the MIME parameter scanner and quoted-string reader to operate on Unicode scalars, so ASCII quotes adjacent to combining marks are recognized as grammar. Added regressions for the forged-boundary case and quoted values beginning with a combining mark.

Sources/SwiftMail/MIME/EMLParser+Parameters.swift
odrobnik: Added RFC 822 comment state, including nesting and quoted-pairs, to parameter splitting and comment removal before attribute/value parsing. Quotes and semicolons inside comments can no longer hide or forge a boundary.

Sources/SwiftMail/MIME/EMLParser+Parameters.swift
odrobnik: Updated RFC 2231 decoding to parse charset and language as separate apostrophe-delimited fields, percent-decode bytes, and decode supported charsets. UTF-8 values with language tags now correctly outrank and replace the fallback.

Sources/SwiftMail/MIME/MIMEHeaderEncoding.swift
odrobnik: Added RFC 2231 continuation emission when a filename parameter would exceed the 998-octet physical-line limit, plus continuation parsing and cleanup. Short parameters remain byte-identical; long Korean filenames fold and round-trip.

Sources/SwiftMail/MIME/MIMEHeaderEncoding.swift
odrobnik: Changed field-body sanitization to preserve RFC 6532 UTF-8 non-ASCII scalars while still removing header-breaking controls such as CR/LF. UTF-8 Content-IDs now serialize and reparse unchanged.

odrobnik:
MissionControl opened #230 from `agent/pr-216-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #230 merges. Thanks for the original contribution!

PR #218, opened 2026-08-30
[WIP] Avoid calling ProcessInfo.processInfo.hostName for privacy fix

nmn:
Calling `ProcessInfo.processInfo.hostName` triggers a privacy confirmation asking for access to local networks. This breaks SMTP for sending emails if you decline the permission when the prompt shows up.

- Replace hostname lookup in EHLO with a privacy-safe [127.0.0.1] default, avoiding local-network access and device-name disclosure.
- Add configurable, validated clientIdentity support, reused before and after STARTTLS.
- Document the new behavior and add tests for defaults, customization, STARTTLS reuse, and command-injection prevention.

Still trying to improve things.

### DIFF
--- a/Sources/SwiftMail/SMTP/Commands/EHLOCommand.swift
+++ b/Sources/SwiftMail/SMTP/Commands/EHLOCommand.swift
@@ -14,17 +14,28 @@ struct EHLOCommand: SMTPCommand {
     /// Timeout in seconds for EHLO command (typically quick to respond)
     let timeoutSeconds: Int = 30
 
-    /// The hostname to use for the EHLO command
-    let hostname: String
+    /// The client identity to use for the EHLO command
+    let clientIdentity: String
 
     /// Initialize a new EHLO command
-    /// - Parameter hostname: The hostname to use for the EHLO command
-    init(hostname: String) {
-        self.hostname = hostname
+    /// - Parameter clientIdentity: The RFC 5321 domain or address literal to use for the EHLO command
+    init(clientIdentity: String) {
+        self.clientIdentity = clientIdentity
     }
 
     /// Convert the command to a string that can be sent to the server
     func toCommandString() -> String {
-        return "EHLO \(hostname)"
+        return "EHLO \(clientIdentity)"
+    }
+
+    /// Reject malformed identities before they can alter the SMTP command stream.
+    func validate() throws {
+        let bytes = clientIdentity.utf8
+        guard (1...255).contains(bytes.count),
+            bytes.allSatisfy({ (33...126).contains($0) }) else {
+            throw SMTPError.commandFailed(
+                "EHLO client identity must be 1...255 printable ASCII characters without whitespace"
+            )
+        }
     }
 }
--- a/Sources/SwiftMail/SMTP/SMTPServer.swift
+++ b/Sources/SwiftMail/SMTP/SMTPServer.swift
@@ -56,11 +56,21 @@ public actor SMTPServer {
 
     // MARK: - Properties
 
+    /// The privacy-safe identity used by default in SMTP `EHLO` commands.
+    ///
+    /// RFC 5321 permits an address literal when a client has no stable,
+    /// fully-qualified hostname. This loopback literal requires no hostname,
+    /// DNS, Bonjour, or local-network lookup.
+    public static let defaultClientIdentity = "[127.0.0.1]"
+
     /** The hostname of the SMTP server */
     let host: String
 
     /** The port number of the SMTP server */
     let port: Int
+
+    /// The domain or address literal sent as the client identity in every `EHLO` command.
+    public let clientIdentity: String
 
     /** The requested SMTP transport security policy */
     let transportSecurity: MailTransportSecurity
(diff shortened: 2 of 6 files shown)


Sources/SwiftMail/SMTP/Commands/EHLOCommand.swift
```diff
@@ -14,17 +14,28 @@ struct EHLOCommand: SMTPCommand {
...
     func toCommandString() -> String {
-        return "EHLO \(hostname)"
+        return "EHLO \(clientIdentity)"
+    }
+
+    /// Reject malformed identities before they can alter the SMTP command stream.
+    func validate() throws {
+        let bytes = clientIdentity.utf8
+        guard (1...255).contains(bytes.count),
+            bytes.allSatisfy({ (33...126).contains($0) }) else {
```
odrobnik, line 35: **MEDIUM** — An app configures `clientIdentity: "client_name"`, which is printable ASCII but contains an underscore forbidden in RFC 5321 domain labels. `validate()` succeeds and SwiftMail sends `EHLO client_name`; an RFC-conforming server can reply 501, so connection establishment fails instead of SwiftMail rejecting the malformed identity under its documented validation contract.

Sources/SwiftMail/SMTP/SMTPServer.swift
```diff
@@ -166,10 +180,12 @@ public actor SMTPServer {
         certificateVerificationPolicy: MailCertificateVerificationPolicy = .fullVerification,
         minimumTLSVersion: MailTLSMinimumVersion = .tlsv12,
         numberOfThreads: Int = 1,
-        submissionTimeouts: SMTPSubmissionTimeouts = SMTPSubmissionTimeouts()
+        submissionTimeouts: SMTPSubmissionTimeouts = SMTPSubmissionTimeouts(),
+        clientIdentity: String = SMTPServer.defaultClientIdentity
     ) {
         self.host = host
         self.port = port
+        self.clientIdentity = clientIdentity
```
odrobnik, line 188: **MEDIUM** — An app configures `clientIdentity: "bad identity"` and calls `connect()`. The TCP connection and greeting complete and `self.channel` is set before EHLO validation throws, but `connect()` does not close or clear the channel. Retrying `connect()` on the same actor overwrites that reference and can leak one active socket per retry.

odrobnik (changes requested):
Request changes for two correctness issues in the new client-identity API.

1. `EHLOCommand.validate()` does not enforce the documented RFC 5321 `Domain / address-literal` contract; it accepts any length-bounded printable non-space ASCII token, so malformed configuration reaches the server and can fail EHLO.
2. Identity validation occurs only during capability fetching, after `connect()` has opened and stored the channel. A rejected identity leaves that channel active, and retries on the same `SMTPServer` can overwrite the reference and leak connections.

The privacy-safe default and consistent pre-/post-STARTTLS identity reuse otherwise trace correctly. Builds and tests were not run, per the review constraint.

- **MEDIUM** `Sources/SwiftMail/SMTP/Commands/EHLOCommand.swift:35`
- **MEDIUM** `Sources/SwiftMail/SMTP/SMTPServer.swift:188`

odrobnik:
Shouldn the default identity be "localhost"?

Sources/SwiftMail/SwiftMail.docc/Articles/GettingStartedWithSMTP.md
```diff
@@ -21,6 +21,30 @@ Common SMTP ports:
...
+    port: 587,
+    clientIdentity: "mail.example.com"
+)
+```
+
+SwiftMail uses the configured value for the initial `EHLO` and again after a
+STARTTLS upgrade. Because the initial command can be sent before encryption,
+do not put private device or user information in this value. The identity must
+be a single printable ASCII domain or address literal as defined by
+[RFC 5321](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.1.1.1).
```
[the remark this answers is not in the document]
odrobnik: Tightened EHLO validation to enforce RFC 5321 Domain syntax and registered IPv4/IPv6 address-literal forms, so the documented contract now matches runtime behavior.

Sources/SwiftMail/SMTP/SMTPServer.swift
```diff
@@ -150,6 +160,10 @@ public actor SMTPServer {
        - certificateVerificationPolicy: The certificate verification policy to use for TLS connections
        - numberOfThreads: The number of threads to use for the event loop group
        - submissionTimeouts: Timeout budgets for the SMTP mail-submission dialogue
+       - clientIdentity: The RFC 5321 domain or address literal sent in `EHLO` commands. The
+         lookup-free default is ``SMTPServer/defaultClientIdentity``. A configured value is
+         sent before TLS as well as after STARTTLS, so it should not contain private device
+         information.
```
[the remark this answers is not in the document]
odrobnik: Tightened EHLO validation to enforce the initializer's documented RFC 5321 Domain/address-literal contract, including label and literal validation.

Sources/SwiftMail/SMTP/Commands/EHLOCommand.swift
odrobnik: The finding was correct. EHLO validation now rejects underscores and other invalid domain-label forms, enforces label length/boundaries, and validates IPv4/IPv6 literals.

Sources/SwiftMail/SMTP/SMTPServer.swift
odrobnik: The finding was correct. connect() now validates clientIdentity before creating a TCP connection, preventing malformed configuration from opening or retaining a channel; an integration regression covers this.

odrobnik:
Kept `[127.0.0.1]` as the default intentionally. RFC 5321 recommends an address literal when the client has no meaningful fully qualified domain name, while `localhost` is a local alias; the literal also makes the no-hostname/no-name-service behavior explicit and avoids disclosing a device name.

odrobnik:
MissionControl opened #229 from `agent/pr-218-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #229 merges. Thanks for the original contribution!

PR #219, opened 2026-09-01
Add validated partial IMAP body part fetching

tabmail-kmyi:
## Status

Ready for review.

## Why

SwiftMail currently fetches an entire MIME part in one IMAP response. A caller cannot bound the response size for a large part, and cannot tell whether a server honored a requested partial range. Raising a parser limit only moves that failure threshold.

## What

- Add `fetchPart(section:of:offset:count:)` for `BODY.PEEK[section]<offset.count>` requests.
- Reuse `FetchMessagePartCommand` and its existing full-part path; the range is optional.
- Validate message identity, section, echoed origin, declared length, received length, duplicate groups, and ambiguous responses before returning bytes.
- Return transfer-encoded bytes unchanged so callers can concatenate chunks before decoding.
- Keep diagnostics bounded while omitting streaming literal fragments from logs.
- Fail immediately and recycle the connection when a malformed streaming declaration proves unread literal bytes remain.

## How

The command uses SwiftNIO IMAP's existing ranged `bodySection` grammar. The existing fetch handler gains request-aware validation only when a range is present, leaving ordinary full-part behavior unchanged. Valid tagged `BAD`, tagged `NO`, no-match, malformed-response, and transport-terminal outcomes remain distinct so callers can make truthful retry decisions.

## Risks

The validation deliberately rejects ambiguous or non-conforming partial responses rather than returning potentially wrong bytes. Such a response also forces a reconnect because its unread literal can no longer be parsed safely. Compliant full and partial fetch behavior is unchanged.

## Verification

- Full Swift package suite: 508 tests across 65 suites passed.
- Focused partial-fetch suite: 13 tests across 3 suites passed.
- Reconstructed a 33 MiB + 257 byte part in 1 MiB chunks with a 4 MiB parser buffer, including zero-byte EOF and sequence-number fetching.
- Covered ignored ranges larger than 32 MiB, wrong origins, malformed literals, UID conflicts, opaque bytes, tagged `BAD`/`NO`, BYE/FATAL, and bounded logger state.
- Strict SwiftLint passed on all changed files.
- Four-angle review gate: architecture, correctness, robustness, and mutation-based test coverage.

## Scope

The patch adds 319 production lines and 666 test lines. It does not overlap the files changed by open PRs #214, #215, #216, or #218.

## Next action

Maintainer review and merge.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Commands/FetchCommands.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Commands/FetchCommands.swift
@@ -101,6 +101,9 @@ struct FetchMessagePartCommand<T: MessageIdentifier>: IMAPTaggedCommand {
     /// The section path to fetch (e.g., [1], [1, 1], [2], etc.)
     let section: Section
 
+    /// Optional encoded-byte range for a partial body fetch.
+    let range: ClosedRange<UInt32>?
+
     /// Custom timeout for this operation
     var timeoutSeconds: Int { return 60 }
 
@@ -108,9 +111,29 @@ struct FetchMessagePartCommand<T: MessageIdentifier>: IMAPTaggedCommand {
     /// - Parameters:
     ///   - identifier: The message identifier to fetch
     ///   - sectionPath: The section path to fetch as an array of integers
-    init(identifier: T, section: Section) {
+    init(identifier: T, section: Section, range: ClosedRange<UInt32>? = nil) {
         self.identifier = identifier
         self.section = section
+        self.range = range
+    }
+
+    func makeHandler(commandTag: String, promise: EventLoopPromise<Data>) -> FetchPartHandler {
+        guard let range else {
+            return FetchPartHandler(commandTag: commandTag, promise: promise)
+        }
+        let expectedIdentifier: PartialFetchIdentifier = T.self == UID.self
+            ? .uid(identifier.value)
+            : .sequenceNumber(identifier.value)
+        return FetchPartHandler(
+            commandTag: commandTag,
+            promise: promise,
+            partialRequest: .init(
+                identifier: expectedIdentifier,
+                section: SectionSpecifier(part: .init(section.components)),
+                offset: Int(range.lowerBound),
+                count: range.count
+            )
+        )
     }
 
     /// Convert to an IMAP tagged command
@@ -124,7 +147,7 @@ struct FetchMessagePartCommand<T: MessageIdentifier>: IMAPTaggedCommand {
         let section = SectionSpecifier(part: part)
 
         let attributes: [FetchAttribute] = [
-            .bodySection(peek: true, section, nil)
+            .bodySection(peek: true, section, range)
         ]
 
         if T.self == UID.self {
--- a/Sources/SwiftMail/IMAP/IMAP/Handler/FetchPartHandler.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Handler/FetchPartHandler.swift
@@ -5,99 +5,241 @@ import Foundation
 @preconcurrency import NIOIMAP
 import NIOIMAPCore
 import NIO
-import NIOConcurrencyHelpers
 
-/// Handler for IMAP FETCH PART command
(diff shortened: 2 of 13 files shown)


Sources/SwiftMail/IMAP/IMAP/Commands/FetchCommands.swift
```diff
@@ -101,16 +101,39 @@ struct FetchMessagePartCommand<T: MessageIdentifier>: IMAPTaggedCommand {
...
         self.identifier = identifier
         self.section = section
+        self.range = range
+    }
+
+    func makeHandler(commandTag: String, promise: EventLoopPromise<Data>) -> FetchPartHandler {
+        guard let range else {
+            return FetchPartHandler(commandTag: commandTag, promise: promise)
+        }
+        let expectedIdentifier: PartialFetchIdentifier = T.self == UID.self
```
odrobnik, line 124: **MEDIUM** — `UID.latest` and `SequenceNumber.latest` encode on the wire as the wildcard `*`, but the handler captures their raw `UInt32.max` value here as an exact expected response identifier. On any nonempty mailbox, `fetchPart(... of: UID.latest, offset: 0, count: 4)` receives the concrete last UID, never matches it, fails with `invalidResponse`, and recycles a valid connection; sequence-number `.latest` fails the same way.

Sources/SwiftMail/IMAP/IMAP/Handler/FetchPartHandler.swift
```diff
@@ -5,99 +5,241 @@ import Foundation
...
+    private func beginBody(kind: StreamingKind, declaredCount: Int, request: PartialFetchRequest) {
+        currentBodyCount += 1
+        guard currentBodyCount == 1 else {
+            rejectUnsafeResponse("multiple body literals")
+            return
+        }
+        guard case .body(let section, let offset) = kind,
+              section == request.section,
+              offset == request.offset else {
+            rejectUnsafeResponse("section or partial origin did not match request")
```
odrobnik, line 193: **MEDIUM** — Request-specific body validation runs before the FETCH group's message identity is filtered. While fetching UID 42/section 1, a valid unsolicited response for another message such as `* 7 FETCH (UID 99 BODY[2] {4}\r\njunk)` reaches this guard, immediately fails for the section/origin mismatch, and disconnects before UID 42's valid response arrives. Identified unrelated body groups must be discarded rather than fail the active command.

odrobnik (changes requested):
Requesting changes for two correctness/interoperability regressions in the new partial-fetch path: wildcard `.latest` identifiers can never match their concrete replies, and a valid unsolicited FETCH body for another message is rejected before the response group is filtered by identity. Both cases turn valid server behavior into `invalidResponse` and recycle a healthy connection.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAP/Commands/FetchCommands.swift:124`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAP/Handler/FetchPartHandler.swift:193`

Sources/SwiftMail/IMAP/IMAP/Commands/FetchCommands.swift
odrobnik: Fixed by representing UID.latest and SequenceNumber.latest as wildcard expectations in the partial-fetch handler, so concrete server response identifiers match. Added command-level tests for both wildcard forms.

Sources/SwiftMail/IMAP/IMAP/Handler/FetchPartHandler.swift
odrobnik: Fixed by classifying each FETCH group as matching, unrelated, or pending before applying request-specific body validation. Identified unrelated body groups are now discarded, with regression coverage for an unsolicited mismatched section followed by the requested response.

odrobnik:
MissionControl opened #228 from `agent/pr-219-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #228 merges. Thanks for the original contribution!

PR #220, opened 2026-09-01
Expose Gmail attributes on named IMAP connections

jdudley:
## Summary

- expose Gmail extension capability state on IMAPNamedConnection
- add fetchGmailAttributes(for:) on named connections, matching the existing IMAPServer API
- preserve UID-keyed results when FETCH responses arrive out of request order

## Motivation

Consumers that select and mutate mailboxes through a named IMAP connection need to fetch Gmail-native metadata on that same session. Kestrel Mail uses this to map selected IMAP UIDs to exact X-GM-MSGID values before permanent deletion, without switching back to the primary connection.

## Validation

- swiftlint lint --strict --no-cache
- swift test --filter FetchGmailAttributesTests (5 tests passed)

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
+++ b/Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
@@ -1,0 +1,34 @@
+import Foundation
+import NIOIMAPCore
+
+extension IMAPNamedConnection {
+    /// Whether this connection advertised Gmail's `X-GM-EXT-1` capability.
+    public var supportsGmailExtensions: Bool {
+        capabilities.contains(.gmailExtensions)
+    }
+
+    /// Fetches Gmail-native attributes for the given UIDs on this named connection.
+    ///
+    /// Requires the `X-GM-EXT-1` capability; other IMAP servers answer with a
+    /// tagged BAD. Gate calls on `Capability.gmailExtensions` being advertised.
+    public func fetchGmailAttributes(
+        for identifierSet: UIDSet
+    ) async throws -> [UID: GmailMessageAttributes] {
+        let command = FetchGmailAttributesCommand(identifierSet: identifierSet)
+        let records = try await executeCommand(command)
+
+        var result: [UID: GmailMessageAttributes] = [:]
+        for record in records {
+            guard let uid = record.uid,
+                  let messageID = record.messageID,
+                  let threadID = record.threadID
+            else { continue }
+            result[uid] = GmailMessageAttributes(
+                messageID: messageID,
+                threadID: threadID,
+                labels: record.labels
+            )
+        }
+        return result
+    }
+}
--- a/Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
+++ b/Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
@@ -4,13 +4,13 @@ import NIOIMAPCore
 extension IMAPNamedConnection {
     /// Whether this connection advertised Gmail's `X-GM-EXT-1` capability.
     public var supportsGmailExtensions: Bool {
-        capabilities.contains(.gmailExtensions)
+        capabilities.containsGmailExtensionsCapability
     }
 
     /// Fetches Gmail-native attributes for the given UIDs on this named connection.
     ///
     /// Requires the `X-GM-EXT-1` capability; other IMAP servers answer with a
-    /// tagged BAD. Gate calls on `Capability.gmailExtensions` being advertised.
+    /// tagged BAD. Gate calls on `supportsGmailExtensions`.
     public func fetchGmailAttributes(
         for identifierSet: UIDSet
     ) async throws -> [UID: GmailMessageAttributes] {
--- a/Tests/SwiftIMAPTests/FetchGmailAttributesTests.swift
+++ b/Tests/SwiftIMAPTests/FetchGmailAttributesTests.swift
@@ -118,7 +118,7 @@ struct FetchGmailAttributesTests {
         var response = channel.allocator.buffer(capacity: 0)
         response.writeString(
             "* 1 FETCH (UID 20 X-GM-MSGID 2222222222222222 X-GM-THRID 3333333333333333 "
-                + "X-GM-LABELS (\\Inbox))\r\n"
+                + "X-GM-LABELS (\\Inbox \"Work\"))\r\n"
                 + "* 2 FETCH (UID 10 X-GM-MSGID 4444444444444444 X-GM-THRID 5555555555555555 "
                 + "X-GM-LABELS (\\Important))\r\n"
                 + "A001 OK UID FETCH completed\r\n"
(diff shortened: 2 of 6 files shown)


Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
```diff
@@ -0,0 +1,34 @@
+import Foundation
+import NIOIMAPCore
+
+extension IMAPNamedConnection {
+    /// Whether this connection advertised Gmail's `X-GM-EXT-1` capability.
+    public var supportsGmailExtensions: Bool {
```
odrobnik, line 6: **MEDIUM** — A caller using IMAPServer.fetchGmailAttributes has no public API to check whether the server advertises X-GM-EXT-1 before sending the command. IMAPServer.capabilities is internal, so the only recourse is catching a tagged BAD response. The PR adds supportsGmailExtensions only to IMAPNamedConnection, creating an asymmetry not present in any of the established supportsUIDPlus / supportsMove pairs.

Tests/SwiftIMAPTests/FetchGmailAttributesTests.swift
```diff
@@ -94,6 +94,42 @@ struct FetchGmailAttributesTests {
...
+            "* 1 FETCH (UID 20 X-GM-MSGID 2222222222222222 X-GM-THRID 3333333333333333 "
+                + "X-GM-LABELS (\\Inbox))\r\n"
+                + "* 2 FETCH (UID 10 X-GM-MSGID 4444444444444444 X-GM-THRID 5555555555555555 "
+                + "X-GM-LABELS (\\Important))\r\n"
+                + "A001 OK UID FETCH completed\r\n"
+        )
+        try await channel.writeInbound(response)
+
+        let result = try await resultTask.value
+        #expect(result[UID(20)]?.messageID == 2_222_222_222_222_222)
```
odrobnik, line 130: **LOW** — testNamedConnectionResultsAreKeyedByReturnedUID only asserts messageID for the two UIDs. A regression where threadID or labels are not populated (or are swapped) in the named-connection code path would pass the test suite undetected.

odrobnik (changes requested):
The implementation logic is correct — `IMAPNamedConnection.fetchGmailAttributes` is a faithful mirror of `IMAPServer.fetchGmailAttributes`, `validate()` fires through `executeCommandBody`, UID-keying works, and the new test exercises the named-connection code path end-to-end.

Two issues need attention before merge:

**1 (MEDIUM) — `IMAPServer` is missing `supportsGmailExtensions`**

Before this PR neither type had the capability gate. This PR adds it to `IMAPNamedConnection` but leaves `IMAPServer` without a symmetric property. Because `IMAPServer.capabilities` is `internal`, there is no public way for a caller of `IMAPServer.fetchGmailAttributes` to gate on `X-GM-EXT-1` being advertised. The documentation on both methods says "Gate calls on `Capability.gmailExtensions` being advertised", but only named-connection callers can now honour that guidance through the public API. A parallel `public var supportsGmailExtensions: Bool` on `IMAPServer` (matching `supportsUIDPlus` / `supportsMove`) would close the gap.

**2 (LOW) — New test only validates `messageID`, not `threadID` or `labels`**

`testResultsAreKeyedByReturnedUIDNotResponseOrder` (the `IMAPServer` counterpart) asserts all three fields for both UIDs. `testNamedConnectionResultsAreKeyedByReturnedUID` checks only `messageID`. A regression in `threadID` or `labels` passthrough via the named-connection code path would go undetected.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift:6`
- **LOW** `Tests/SwiftIMAPTests/FetchGmailAttributesTests.swift:129`

Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
jdudley: Fixed in 29885ac: added `public var supportsGmailExtensions: Bool` on `IMAPServer` (in `IMAPServer+Gmail.swift`, alongside `fetchGmailAttributes`), reading the primary connection capabilities the same way `supportsUIDPlus` and `supportsMove` do.

Tests/SwiftIMAPTests/FetchGmailAttributesTests.swift
jdudley: Fixed in 29885ac: `testNamedConnectionResultsAreKeyedByReturnedUID` now uses a mixed label set and asserts `messageID`, `threadID`, and `labels` for both UIDs, matching the `IMAPServer` counterpart.

Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
```diff
@@ -0,0 +1,34 @@
...
+        capabilities.contains(.gmailExtensions)
```
odrobnik: **MEDIUM** — This accessor (and the identical new IMAPServer accessor) relies on synthesized, case-sensitive `Capability` equality. NIOIMAPCore preserves the server's spelling, so a valid CAPABILITY response containing `x-gm-ext-1` or `X-Gm-Ext-1` makes `supportsGmailExtensions` return false. A caller following the documented gate then skips `fetchGmailAttributes` and cannot map the selected UIDs to stable Gmail message IDs before permanent deletion, despite the server advertising support. Generalize the existing case-insensitive bare-capability helper used for MOVE/UIDPLUS and use it on both surfaces.

odrobnik (changes requested):
Request changes for one interoperability bug. Both new `supportsGmailExtensions` accessors use case-sensitive `Capability` set membership even though IMAP capability names are case-insensitive and the pinned NIOIMAPCore parser preserves wire spelling. A compliant server or proxy advertising `x-gm-ext-1` is therefore reported unsupported, so callers following the documented gate skip the Gmail-attribute fetch. Reuse the existing case-insensitive bare-capability matcher used for MOVE/UIDPLUS on both connection surfaces. The earlier API-symmetry and field-assertion findings are otherwise addressed.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift:7`

Sources/SwiftMail/IMAP/IMAPNamedConnection+Gmail.swift
jdudley: Fixed in f535abe: the bare-capability matcher behind `containsMoveCapability`/`containsUIDPlusCapability` is now shared (`containsBareCapability(named:)`), `containsGmailExtensionsCapability` joins it, and both `IMAPServer.supportsGmailExtensions` and `IMAPNamedConnection.supportsGmailExtensions` read it. `GmailExtensionsCapabilityTests` pins the canonical, `x-gm-ext-1`, and `X-Gm-Ext-1` spellings on both surfaces alongside the unsupported case. Doc comments on both `fetchGmailAttributes` now point at `supportsGmailExtensions` as the gate.

odrobnik (approved):
Approved at f535abe. The prior case-insensitive X-GM-EXT-1 finding is fixed by the shared bare-capability matcher on both IMAPServer and IMAPNamedConnection, with canonical, lowercase, and mixed-case coverage. The earlier API-symmetry and complete Gmail-field assertion findings also remain addressed. Full review found no new issues. Builds and tests were not run, per the read-only review instruction.

👍 No findings.

PR #221, opened 2026-09-05
Judge the SASL mechanism after the channel is prepared, not against a cleared snapshot

jdudley:
## Problem

`authenticatePlainBody` and `authenticateXOAUTH2Body` check the cached `capabilities` set for the mechanism **before** calling `prepareAuthenticationChannel`. `disconnectBody` clears that set together with the channel, and a recycle after a timeout or a buffered BYE goes through `disconnectBody`. The next authentication on that connection therefore throws `unsupportedAuthMechanism("XOAUTH2 not advertised by server")` without contacting the server, even though the reconnect that would refresh the snapshot is one line further down inside `prepareAuthenticationChannel`. A named connection or IDLE connection that was dropped once cannot recover on its own: every `ensureAuthenticated` fails the same way.

Observed against Gmail on 2026-09-04: a slow AUTHENTICATE hit the 10 s timeout, the connection was recycled, and every later use of that connection reported "XOAUTH2 not advertised by server" for half an hour while Gmail was advertising `AUTH=XOAUTH2` the whole time.

## Change

- Run the mechanism guard after `prepareAuthenticationChannel`, so a nil channel reconnects (and refreshes capabilities) first.
- When the snapshot is empty on a live channel, refresh it with a `CAPABILITY` command (`useCommandBody: true`, the command queue is already held) before deciding.
- A server that genuinely does not advertise the mechanism is still rejected with the same `unsupportedAuthMechanism` error.

## Tests

`AuthenticationCapabilityGuardTests` (testing channel, no network):
- an empty snapshot triggers a `CAPABILITY` refresh before `AUTHENTICATE` for XOAUTH2 and for PLAIN;
- a refreshed snapshot that lacks the mechanism is still rejected;
- a disconnected connection attempts the transport instead of judging the mechanism.

`swiftlint lint --strict` is clean; `ClientIdentificationReplayTests`, `IMAPNamedConnectionTests`, `XOAUTH2AuthenticationHandlerTests`, and `IMAPConnectionTLSModeTests` still pass.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift
@@ -31,10 +31,9 @@ extension IMAPConnection {
 
     func authenticatePlainBody(username: String, password: String) async throws {
         let mechanism = AuthenticationMechanism("PLAIN")
-        let plainCapability = Capability.authenticate(mechanism)
-
-        let channel = try await prepareAuthenticationChannel(operation: "PLAIN authenticate")
-        try await requireAdvertisedAuthenticationMechanism(plainCapability, name: "PLAIN")
+        let channel = try await prepareAuthenticationChannel(
+            operation: "PLAIN authenticate", advertising: .authenticate(mechanism), name: "PLAIN"
+        )
         let tag = generateCommandTag()
 
         let handlerPromise = channel.eventLoop.makePromise(of: [Capability].self)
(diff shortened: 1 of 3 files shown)


Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift
```diff
@@ -146,6 +143,26 @@ extension IMAPConnection {
...
+    /// with the channel, and the reconnect refreshes it. Checking the stale
+    /// empty set first turned every re-authentication on a dropped connection
+    /// into "not advertised by server" without ever contacting the server. When
+    /// the snapshot is empty on a live channel, ask the server before deciding.
+    private func requireAdvertisedAuthenticationMechanism(
+        _ capability: Capability,
+        name: String
+    ) async throws {
+        if capabilities.isEmpty {
+            try await refreshCapabilities(using: [], useCommandBody: true)
```
odrobnik: **MEDIUM** — On an empty capability snapshot, if the server sends BYE or the socket becomes inactive after `prepareAuthenticationChannel` returns but before this refresh starts, `executeCommandBody(CapabilityCommand())` can disconnect the captured channel, reconnect `self.channel`, and successfully refresh capabilities on the replacement. Both PLAIN and XOAUTH2 then continue with their previously captured closed channel, so handler installation or AUTHENTICATE fails; because that is a `ChannelError`, the authentication failure path calls `disconnectBody()` and closes the new healthy connection too. A transient drop during the new refresh therefore still makes transparent re-authentication fail and leaves the connection disconnected.

odrobnik (changes requested):
Request changes: the new empty-snapshot refresh can replace the connection after both authentication methods have already cached the channel they intend to use. `refreshCapabilities(using:useCommandBody:)` enters `executeCommandBody`, which can recycle a newly buffered BYE or clear an inactive channel and reconnect. Authentication then installs its handler and writes on the stale channel; the resulting transport failure also runs `disconnectBody()` against `self.channel`, closing the fresh replacement. Re-read/return the active channel after the refresh (or otherwise keep channel selection and validation atomic). The new fixed-channel tests do not cover this replacement race.

- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift:159`

Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift
jdudley: Fixed in 4273b37. `prepareAuthenticationChannel(operation:advertising:name:)` now owns the whole step: prepare the channel, refresh an empty snapshot, prepare again and return the channel that is live after the refresh, then judge the mechanism. A second refresh covers a reconnect during the first; a snapshot still empty after that is a connection failure rather than a mechanism verdict. New test `refreshThatReplacedTheChannelAuthenticatesOnTheReplacement` drives the refresh to close the original channel and install a replacement, then asserts AUTHENTICATE is written on the replacement and nothing reaches the stale channel.

odrobnik (approved):
Adversarial review complete at 4273b3738358042d68abb2f24dac164985c08008. The authentication guard now refreshes an empty live capability snapshot, re-prepares the transport after refresh-driven replacement, and evaluates the mechanism against the active channel's capabilities. I found no correctness, API-contract, concurrency, or unsafe-failure regressions in the changed implementation or tests.

👍 No findings.

PR #222, opened 2026-09-05
Send CHARSET UTF-8 with SEARCH when a criterion carries non-ASCII text

jdudley:
## Summary

`SEARCH` and `UID SEARCH` never carried a `CHARSET` argument. RFC 3501 §6.4.4 assumes US-ASCII when it is omitted, and NIO IMAP sends any search string that is not short printable ASCII as a literal, so a subject with an em dash, a curly quote, or an accent went out as an 8-bit literal in an ASCII-assumed command. Gmail rejects that outright:

```
A042 UID SEARCH RETURN (COUNT MIN MAX ALL) SUBJECT {26}
A042 BAD Could not parse command
```

Observed from Kestrel Mail against imap.gmail.com on 2026-09-04/05: every remote subject search whose subject carried non-ASCII text failed this way (eight rejections in sixteen hours, plus one `BAD Unknown command <literal fragment>` where Gmail's parser kept reading).

## Change

- `SearchCriteria.requiresUTF8Charset` walks a criterion (through `and`/`or`/`not`) and reports whether any string payload (`bcc`, `body`, `cc`, `from`, `header`, `subject`, `text`, `to`) contains non-ASCII bytes.
- `SearchCriteria.searchCharset(for:)` returns `"UTF-8"` when any criterion needs it, otherwise `nil`.
- `SearchCommand` and `ExtendedSearchCommand` pass that value as `charset:` to NIO's `.search` / `.uidSearch`. ASCII-only searches keep their existing wire shape; non-ASCII searches become `UID SEARCH [RETURN (...)] CHARSET UTF-8 <keys>`, which is the RFC 3501 / RFC 4731 search-program order NIO already emits.
- `SORT` was already sending `sortCharset` and is unchanged.

## Tests

`SearchCharsetTests` renders both commands through `IMAPClientHandler` on a testing channel, answering the synchronizing-literal continuation the way a server would, and asserts:

- non-ASCII subject → `... RETURN (COUNT MIN MAX ALL) CHARSET UTF-8 ... SUBJECT {26}\r\nMorning brief — Saturday`
- ASCII criteria → no `CHARSET`, subject still quoted
- nested `OR FROM ... TEXT {5}\r\ncafé` → `CHARSET UTF-8` on plain `SearchCommand`
- detection covers `header`, `not`, `and`, and leaves `keyword` alone

`swift test --filter 'SearchCharsetTests|WithinSearchTests|ExtendedSearchHandlerTests'` passes (24 tests); `swiftlint lint --strict` is clean.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAP/Commands/ExtendedSearchCommand.swift
+++ b/Sources/SwiftMail/IMAP/IMAP/Commands/ExtendedSearchCommand.swift
@@ -76,6 +76,7 @@ struct ExtendedSearchCommand<T: MessageIdentifier>: IMAPTaggedCommand, Sendable 
         }
 
         let key = SearchKey.and(nioCriteria)
+        let charset = SearchCriteria.searchCharset(for: criteria)
 
         let returnOptions: [SearchReturnOption]
         if useEsearch {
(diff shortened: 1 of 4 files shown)


odrobnik (approved):
Reviewed the exact head f18398a adversarially across both SEARCH command paths, all SearchCriteria cases and nesting, public call sites, NIOIMAP 0.3.0 encoding behavior (including ESEARCH ordering and UTF8=ACCEPT suppression), and the added wire-format tests. The implementation matches the PR requirements, preserves ASCII-only wire output, and introduces no correctness or protocol regressions found.

👍 No findings.

PR #223, opened 2026-09-06
Require a live channel for an authenticated IMAP session

jdudley:
## Summary

`IMAPConnection.isAuthenticated` now requires a live channel, so a named connection whose transport the peer reset re-authenticates before its next command instead of sending that command on a fresh, unauthenticated socket.

## The failure

Observed against Gmail on 2026-09-05: the server reset every socket of an account at once. A named connection (`IMAPServer.connection(named:)`) that had been idle then ran `UID SEARCH`:

1. `IMAPNamedConnection.executeCommand` calls `ensureAuthenticated()`, which returns early because `connection.isAuthenticated` is still `true`: a peer reset does not clear `isSessionAuthenticated`, only `disconnectBody()` and `clearInvalidChannel()` do, and neither has run yet.
2. `IMAPConnection.executeCommandBody` then calls `clearInvalidChannel()`, finds the channel nil, and re-establishes the transport with `connectBody()`.
3. The SEARCH goes out on the new transport before any AUTHENTICATE. Gmail answers `BAD Unknown command <session-id>`, and the connection's own log reports the state that explains it: `channelActive=true authenticated=false`.

`IMAPServer.ensurePrimaryConnectionAuthenticated` has the same shape for the primary connection, so it benefits as well.

## The fix

```swift
var isAuthenticated: Bool {
    isSessionAuthenticated && isConnected
}
```

An authenticated session cannot outlive its channel. Both re-authentication checks (`IMAPNamedConnection.ensureAuthenticated`, `IMAPServer.ensurePrimaryConnectionAuthenticated`) now run `authenticate(on:)` when the channel is gone, and `executeCommandBody` reopens the transport under the AUTHENTICATE command rather than under the user's command. No other reader of `isAuthenticated` changes meaning: `identify` already pairs it with `isConnected`.

## Test

`IMAPNamedConnectionTests.ensureAuthenticatedReauthenticatesWhenTheSessionFlagOutlivesTheChannel` builds a connection with `isSessionAuthenticated = true` and no channel, asserts the named connection reports unauthenticated, and asserts `ensureAuthenticated()` invokes the authenticate closure once. On `main` the closure is never called.

`swift test` passes; `swiftlint lint --strict` is clean on the changed files.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -36,6 +36,18 @@ final class IMAPConnection {
     let responseBuffer = UntaggedResponseBuffer()
     var startTLSUpgradeOverrideForTesting: (() async throws -> Void)?
     var capabilityRefreshOverrideForTesting: (() async throws -> Void)?
+    var connectOverrideForTesting: (() async throws -> Void)?
+    /// Re-authenticates this connection inside the command queue after the command path
+    /// or an IDLE start had to reopen the transport for a session that was authenticated.
+    /// Installed by `IMAPServer` on every connection it authenticates; nil until then.
+    var reauthenticateAfterReconnect: (@Sendable (IMAPConnection) async throws -> Void)?
+    /// Set when an authenticated session's channel is lost (peer reset, buffered BYE,
+    /// timeout recycle) rather than ended on purpose. Cleared by the next successful
+    /// authentication and by an explicit `disconnect()`.
+    var lostAuthenticatedSession = false
+    /// True while LOGIN, AUTHENTICATE PLAIN, or AUTHENTICATE XOAUTH2 runs, so the
+    /// capability refresh inside them cannot trigger a nested re-authentication.
+    var authenticationInProgress = false
 
     let logger: Logging.Logger
     let duplexLogger: IMAPLogger
@@ -250,6 +262,10 @@ final class IMAPConnection {
         self.capabilityRefreshOverrideForTesting = refresh
     }
 
+    func replaceConnectForTesting(_ connect: (() async throws -> Void)?) {
+        self.connectOverrideForTesting = connect
+    }
+
     func connect() async throws {
         try await commandQueue.run { [self] in
             try await self.connectBody()
@@ -265,6 +281,8 @@ final class IMAPConnection {
     func disconnect() async throws {
         try await commandQueue.run { [self] in
             try await self.disconnectBody()
+            // Ending the session on purpose is not losing it.
+            self.lostAuthenticatedSession = false
         }
     }
 }
--- a/Sources/SwiftMail/IMAP/IMAPConnection.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection.swift
@@ -213,8 +213,12 @@ final class IMAPConnection {
         namespaces
     }
 
+    /// An authenticated session needs a live channel. A transport the peer reset leaves
+    /// `isSessionAuthenticated` set until the next command notices the dead channel, and a
+    /// caller that trusted the flag alone then sent an authenticated-state command on the
+    /// fresh transport `executeCommandBody` reopens (Gmail answers `BAD Unknown command`).
     var isAuthenticated: Bool {
-        isSessionAuthenticated
+        isSessionAuthenticated && isConnected
     }
 
     var identifier: String {
--- a/Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift
+++ b/Sources/SwiftMail/IMAP/IMAPConnection+Authentication.swift
@@ -4,32 +4,15 @@ import NIOIMAPCore
 import NIO
 
 extension IMAPConnection {
-    func login(username: String, password: String) async throws {
-        let command = LoginCommand(username: username, password: password)
-        let loginCapabilities = try await executeCommand(command)
-        isSessionAuthenticated = true
-        try await refreshCapabilities(using: loginCapabilities)
-        await fetchNamespacesIfSupported(useCommandBody: false)
-    }
-
     /// Authenticate using AUTHENTICATE PLAIN (RFC 4616) with optional SASL-IR (RFC 4959).
     ///
     /// When the server advertises `SASL-IR`, the credentials are sent inline with the
     /// AUTHENTICATE command (saving a round trip). Otherwise falls back to the standard
-    /// continuation-based exchange.
-    func authenticatePlain(username: String, password: String) async throws {
-        try await commandQueue.run { [self] in
-            try await self.authenticatePlainBody(username: username, password: password)
-        }
-    }
-
-    func authenticateXOAUTH2(email: String, accessToken: String) async throws {
-        try await commandQueue.run { [self] in
-            try await self.authenticateXOAUTH2Body(email: email, accessToken: accessToken)
-        }
-    }
-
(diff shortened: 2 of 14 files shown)


Sources/SwiftMail/IMAP/IMAPConnection.swift
```diff
@@ -213,8 +213,12 @@ final class IMAPConnection {
         namespaces
     }
 
+    /// An authenticated session needs a live channel. A transport the peer reset leaves
+    /// `isSessionAuthenticated` set until the next command notices the dead channel, and a
+    /// caller that trusted the flag alone then sent an authenticated-state command on the
+    /// fresh transport `executeCommandBody` reopens (Gmail answers `BAD Unknown command`).
     var isAuthenticated: Bool {
-        isSessionAuthenticated
+        isSessionAuthenticated && isConnected
```
odrobnik, line 234: **HIGH** — When a PLAIN or XOAUTH2 session loses its channel, this predicate now sends the caller into reauthentication, but authenticatePlainBody/authenticateXOAUTH2Body check the cached capabilities before prepareAuthenticationChannel reconnects and refreshes them. If the prior post-authentication snapshot omits AUTH=PLAIN/XOAUTH2 (as servers commonly do once authenticated), the next named or primary command throws unsupportedAuthMechanism without opening a replacement transport, so the stale-session recovery this change promises still fails.

odrobnik, line 234: **HIGH** — The liveness/authentication check occurs before the caller later acquires IMAPConnection.commandQueue for its command. If the peer resets after this expression returns true, or an earlier queued command receives BYE and disconnects the channel while this command waits, executeCommandBody clears/reopens the transport and sends the already-approved authenticated-state command without AUTHENTICATE, reproducing the Gmail BAD failure.

odrobnik, line 234: **MEDIUM** — IMAPServer is reentrant across await and its primary reauthentication path has no single-flight coordination. After a disconnect, two concurrent primary operations can both observe false and enqueue authentication; after the first establishes a session, the second attempts LOGIN/AUTHENTICATE again on that authenticated session, causing one valid operation to fail and potentially invoking an OAuth token provider twice.

odrobnik (changes requested):
Requesting changes: requiring a live channel improves stale-session reporting, but it does not yet make reconnect recovery reliable. PLAIN/XOAUTH2 can reject the previous capability snapshot before reconnecting, authentication and the following command are not atomic, concurrent primary callers can start duplicate authentication attempts, and the classic primary IDLE path bypasses reauthentication. The counter-only test does not exercise these transport and authentication paths.

One review with 4 findings — those on changed lines are detailed in inline comments:

- **HIGH** `Sources/SwiftMail/IMAP/IMAPConnection.swift:221`
- **HIGH** `Sources/SwiftMail/IMAP/IMAPConnection.swift:221`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPConnection.swift:221`
- **MEDIUM** `Sources/SwiftMail/IMAP/IMAPServer+Idle.swift` — Public IMAPServer.idle() calls primaryConnection.idle() directly and never consults ensurePrimaryConnectionAuthenticated or the changed predicate. After an authenticated primary channel drops while the cached IDLE capability remains, startIdleSession clears/reconnects the transport and writes IDLE on the fresh unauthenticated session, so classic IDLE still fails under the reset scenario this PR is intended to repair.

Sources/SwiftMail/IMAP/IMAPConnection.swift
jdudley: Fixed by merging main: the guard that judged the cached snapshot before preparing the channel is #221, which landed after this branch was cut. The branch now carries it, and the recovery path added in 63e6a85 runs those same guarded authenticate bodies.

jdudley: Fixed in 63e6a85. The liveness check is no longer what recovery relies on. The connection records a lost authenticated session when its channel dies or is recycled under an authenticated session, and when executeCommandBody has to reopen the transport for one it re-authenticates on that same socket inside the same queue turn, before the command is written. Test commandThatReopensTheTransportReauthenticatesFirst closes the channel under an authenticated session and asserts the re-authentication runs before NOOP reaches the replacement.

jdudley: Fixed in 63e6a85. ensurePrimaryConnectionAuthenticated now shares one in-flight authentication: a caller that finds one running awaits it instead of starting its own. The in-queue recovery is single-flight by construction as well, since the queue serializes the reconnect and the re-authentication. Test concurrentPrimaryCallersShareOneAuthentication drives two callers and asserts exactly one AUTHENTICATE.

jdudley:
Regarding the fourth finding (IMAPServer+Idle.swift, classic IDLE bypasses re-authentication): fixed in 63e6a85. startIdleSession runs the same in-queue recovery right after its reconnect, so IDLE on a reopened transport is preceded by AUTHENTICATE; the hook is installed on the primary connection (which the public idle() uses) and on every named and IDLE connection the server creates. Test idleStartThatReopensTheTransportReauthenticatesFirst covers it.

odrobnik:
@jdudley Copilot says this to the macOS CI test failure:

The job is failing because the macOS test run is hitting Swift 6 concurrency warnings that are treated as errors in this setup:

- `Tests/SwiftIMAPTests/ReauthenticationAfterReconnectTests.swift`
- repeated on lines `31`, `49`, `67`, and `85`

The issue is this pattern inside `async` test methods:

```swift
defer { try? group.syncShutdownGracefully() }
```

`syncShutdownGracefully()` is unavailable from asynchronous contexts in Swift 6, so these warnings stop the build.

## Fix

Convert the shutdown to async, or move shutdown outside the async context.

### Recommended code change
Replace each `defer` with an async-friendly shutdown helper, for example:

```swift
let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
defer {
    Task {
        try? await group.shutdownGracefully()
    }
}
```

If you need deterministic cleanup before the test exits, prefer an explicit `defer`-safe async teardown pattern, such as:

```swift
let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
defer {
    group.shutdownGracefully { _ in }
}
```

if the API in your NIO version supports the callback-based shutdown from async contexts.

## Best targeted fix in this file

Since this is repeated in four tests, add a small helper in the test suite and use it consistently:

```swift
private func makeGroup() -> MultiThreadedEventLoopGroup {
    MultiThreadedEventLoopGroup(numberOfThreads: 1)
}
```

Then handle shutdown in a non-blocking way compatible with Swift 6, rather than calling the synchronous variant directly inside `async` tests.

## Why this is the failing job

The log shows the compiler complaining at exactly those lines, and the job later exits with code 137 after the test run stalls. The concurrency warning is the actionable failure; the stall is a side effect of the test process not completing cleanly.

## Focused solution

Update `ReauthenticationAfterReconnectTests.swift` to remove all `group.syncShutdownGracefully()` calls from async test contexts. That should unblock the Swift 6 build and prevent the test runner from hanging.

If you want, I can also help draft the exact patch for this test file.

odrobnik:
MissionControl opened #224 from `agent/pr-223-supersede` to carry this work forward — it keeps the commits from this pull/merge request and adds the fixes requested in review. This request can be closed once #224 merges. Thanks for the original contribution!

PR #227, opened 2026-09-07
Surface BADCHARSET from SEARCH as a typed error and keep response codes in search failures

jdudley:
## Summary

Both search handlers dropped the response code of a NO/BAD and reported only its text. Two things were lost:

- `NO [BADCHARSET (US-ASCII)] The specified charset is not supported.` is how Exchange Online answers every `SEARCH ... CHARSET UTF-8`, which #222 now sends for non-ASCII criteria. RFC 3501 §7.1 (and RFC 9051 §7.1) define BADCHARSET for exactly this, with an optional list of the charsets the server does support. Callers saw a generic `commandFailed` and could not tell "this server cannot search this text" from any other failure.
- Every other code (`[NONEXISTENT]`, `[THROTTLED]`, `[LIMIT]`, …) disappeared from the message.

## Changes

- `IMAPError.searchCharsetNotSupported(supportedCharsets:reason:)`, thrown by `SearchHandler` and `ExtendedSearchHandler` when the rejection carries `[BADCHARSET ...]`. The list is the server's; empty when it named none. Description, failure reason, and recovery suggestion follow the existing cases.
- `IMAPError.searchRejected(operation:status:responseText:)` builds the error for both handlers, replacing eight hand-formatted strings. Any other rejection keeps its message shape (`Extended search failed: NO ...`) with the response code rendered in wire form through NIO's `ResponseText` description, so a rejection without a code is worded exactly as before.

No retry is attempted here: a raw non-ASCII literal without CHARSET is what Gmail rejects (#222), so the right follow-up on a BADCHARSET is the caller's to decide, and this PR gives it the information to do so.

## Tests

`SearchBadCharsetTests` (6) drive both handlers through the IMAP client parser with the wire lines: BADCHARSET with a list, without a list, untagged before the tagged line, another code (`[NONEXISTENT]`) kept in the text, a code-less `BAD Could not parse command` worded as before, and the description strings. Full suite: 523 tests pass. `swiftlint lint --strict` is clean.

## Context

Seen live on 2026-09-07 from Kestrel Mail against `outlook.office365.com`: the first two Outlook context searches with a non-ASCII subject after repinning to #224 failed this way.

### DIFF
--- a/Sources/SwiftMail/IMAP/IMAPError.swift
+++ b/Sources/SwiftMail/IMAP/IMAPError.swift
@@ -204,7 +204,8 @@ extension IMAPError: LocalizedError {
             case .malformedCopyUIDAfterTaggedOK:
                 return "Do not retry. The command completed; refresh the affected mailboxes before continuing."
             case .searchCharsetNotSupported:
-                return "Retry with a charset the server supports, or limit the search text to US-ASCII."
+                return "Retry with a charset the server supports: US-ASCII-only search text for SEARCH, "
+                    + "or a supported sortCharset for SORT."
             default:
                 return "Check the error details and try again."
         }
@@ -216,7 +217,14 @@ extension IMAPError {
     /// response carries that code, otherwise `commandFailed` with the response code kept in wire
     /// form, so `[NONEXISTENT]`, `[THROTTLED]` and the like stay readable in the message.
     static func searchRejected(operation: String, status: String, responseText: ResponseText) -> IMAPError {
-        let reason = "\(operation) failed: \(status) \(responseText.debugDescription)"
+        // Rendered by hand rather than through ResponseText's description: NIO accepts a NO or
+        // BAD with no text (iCloud, Oracle) and the description substitutes a placeholder space,
+        // which would turn the established "NO " wording of a code-less rejection into "NO  ".
+        var reason = "\(operation) failed: \(status)"
+        if let code = responseText.code {
+            reason += " [\(code.debugDescription)]"
+        }
+        reason += " \(responseText.text)"
         if case .badCharset(let supportedCharsets)? = responseText.code {
             return .searchCharsetNotSupported(supportedCharsets: supportedCharsets, reason: reason)
         }
--- a/Sources/SwiftMail/IMAP/IMAPError.swift
+++ b/Sources/SwiftMail/IMAP/IMAPError.swift
@@ -2,6 +2,7 @@
 // Custom IMAP errors
 
 import Foundation
+import NIOIMAPCore
 
 /// Errors that can occur during IMAP operations
 public enum IMAPError: Error {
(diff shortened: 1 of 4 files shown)


Sources/SwiftMail/IMAP/IMAPError.swift
```diff
@@ -190,13 +203,26 @@ extension IMAPError: LocalizedError {
...
         }
     }
 }
 
 extension IMAPError {
+    /// The error for a rejected SEARCH, UID SEARCH, or SORT: the typed BADCHARSET case when the
+    /// response carries that code, otherwise `commandFailed` with the response code kept in wire
+    /// form, so `[NONEXISTENT]`, `[THROTTLED]` and the like stay readable in the message.
+    static func searchRejected(operation: String, status: String, responseText: ResponseText) -> IMAPError {
+        let reason = "\(operation) failed: \(status) \(responseText.debugDescription)"
```
odrobnik: **LOW** — Preserve empty-text rejection wording instead of routing it through ResponseText.debugDescription. NIO deliberately accepts tagged or untagged `NO`/`BAD` responses with no response text (for iCloud/Oracle compatibility) and represents them as `text == ""`; its debugDescription synthesizes a single placeholder space. Because this interpolation already inserts a space after the status, `A1 NO` now produces `Search failed: NO  ` rather than the previous `Search failed: NO `, violating the PR's stated compatibility guarantee for code-less rejections and changing the public associated reason string on real supported server responses.

Sources/SwiftMail/IMAP/IMAPError.swift
```diff
@@ -190,13 +203,26 @@ extension IMAPError: LocalizedError {
                 return "Do not retry. Refresh both mailboxes and reconcile their current state first."
             case .malformedCopyUIDAfterTaggedOK:
                 return "Do not retry. The command completed; refresh the affected mailboxes before continuing."
+            case .searchCharsetNotSupported:
```
odrobnik, line 206: **LOW** — Qualify the US-ASCII recovery alternative so it is not offered for SORT/UID SORT. Those commands always carry the separate `sortCharset` argument (default `UTF-8`) regardless of whether the search criteria are ASCII. If a SORT server rejects UTF-8 with BADCHARSET and a caller follows this suggestion by only limiting the search text to US-ASCII, the retry still emits `SORT ... UTF-8 ...` and fails identically; the caller must instead change `sortCharset` to a supported value.

odrobnik (changes requested):
Requesting changes for two low-severity API/diagnostic edge cases. The BADCHARSET classification and response-code preservation otherwise trace correctly across SEARCH, UID SEARCH, SORT, and UID SORT.

- **LOW** `Sources/SwiftMail/IMAP/IMAPError.swift:219`
- **LOW** `Sources/SwiftMail/IMAP/IMAPError.swift:206`

Sources/SwiftMail/IMAP/IMAPError.swift
jdudley: Accurate. Fixed in e34d33a: the reason is built by hand (status, then `[code]` when present, then the text) instead of going through `ResponseText.debugDescription`, so `A1 NO` renders `Search failed: NO ` exactly as before. `rejectionsWithoutTextKeepTheirWording` drives that line through the parser, and `badCharsetWithoutTextCarriesTheCodeAlone` covers a code with empty text.

Sources/SwiftMail/IMAP/IMAPError.swift
jdudley: Accurate. Fixed in e34d33a: the suggestion now reads "Retry with a charset the server supports: US-ASCII-only search text for SEARCH, or a supported sortCharset for SORT."

odrobnik (approved):
Approved at e34d33a. The BADCHARSET typed-error classification and response-code preservation trace correctly through SEARCH, UID SEARCH, SORT, and UID SORT. The follow-up preserves code-less rejection wording and gives operation-appropriate recovery guidance. No remaining correctness, concurrency, failure-safety, or API-contract findings.

👍 No findings.

PR #233, opened 2026-09-10
Read RFC 2231 continuations in one pass and honor every charset the platform names

tabmail-kmyi:
Two regressions from #230, both in how `EMLParser` reads RFC 2231 extended parameters. Neither touches the parameter-boundary hardening that commit added.

### Continuation sections cost quadratic work

`extendedContinuation` found each numbered section with `extractHeaderParam`, and that function tokenizes the whole header on every call. A sender who writes N legal `filename*N*=` sections therefore made the parser do work proportional to N². Measured with `swiftc -O -wmo` on the exact function source:

| sections | header bytes | before #230 | #230 |
|---:|---:|---:|---:|
| 1,024 | 17,339 | 0.012 s | 0.50 s |
| 2,048 | 35,771 | 0.020 s | 1.97 s |
| 4,096 | 72,635 | 0.031 s | 8.08 s |
| 8,192 | 146,363 | 0.056 s | did not finish in 25 s |

The full `EMLParser.parse` reaches this through an ordinary multipart body with folded header lines, so a small received attachment can hold a parse for seconds.

The header is now tokenized once into `parameters(of:)`, and both `extractHeaderParam` and the continuation collector read from that list. The collector also applies RFC 2231 §3 as written: sections count in decimal from 0, no leading zeroes, no gaps, and where a section repeats the first occurrence stands, matching `extractHeaderParam`.

### Blank and non-whitelisted charsets discarded the filename

`decodeExtendedBytes` accepted only `utf-8`, `us-ascii` and `iso-8859-1` and returned nil for every other label, including the blank one. RFC 2231 §4 says "it is perfectly permissible to leave either the character set or language field blank" with the `'` delimiters present, so `filename*=''invoice.pdf` is a legal spelling of `invoice.pdf`, and `windows-1252` or `iso-8859-15` labels are ordinary real-world mail. All of these lost the name and fell through to a generated one.

A blank charset now decodes as UTF-8, which reads every US-ASCII value unchanged and still rejects bytes that are not text. Any other label resolves through `String.Encoding(ianaCharsetName:)`, the same lookup the encoded-word decoder already uses. A label the platform cannot name still yields nil, and `extractFilename` falls through to the literal spelling.

### Layout

The RFC 2231 helpers move to `EMLParser+RFC2231.swift` so `EMLParser+Parameters.swift` stays under the lint file-length limit. Nothing changes in the move beyond the two fixes above.

### Verification

- `swift test`: 624 tests in 77 suites pass (619 existing plus the five below).
- `swiftlint lint --strict`: 0 violations.
- New tests in `MIMEParameterParsingTests`: blank charset; `windows-1252`, `iso-8859-15` and `ISO-8859-1` labels; unknown charset falling back to the literal spelling; section-numbering rules; and a 4,096-section header that must decode in under two seconds. On the parent commit the blank-charset, platform-charset and linear-work tests fail (the last after 34 s in a debug build); the other two pin behaviour that already held.

https://claude.ai/code/session_0132BGdMBGzmQwA6jkB3rPma

### DIFF
--- a/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
@@ -60,16 +60,32 @@ extension EMLParser {
         return String(bytes: bytes, encoding: encoding)
     }
 
-    /// Whether a charset label names UTF-8 itself (or its US-ASCII subset).
+    /// Whether a charset label names UTF-8 itself.
     ///
     /// On platforms without CoreFoundation, `String.Encoding(ianaCharsetName:)`
     /// resolves legacy charsets it has no converter for (GBK, Big5, EUC-KR,
     /// KOI8-R, …) to `.utf8` as a best-effort placeholder. Here a wrong
     /// decode is worse than none, because `nil` lets the sender's literal
-    /// spelling win, so a UTF-8 result is trusted only for a UTF-8 label.
+    /// spelling win, so a `.utf8` result is trusted only for a label that
+    /// really names UTF-8. The label is normalized the way the resolver
+    /// normalizes it before lookup — trimmed, unquoted, lowercased, `_` read
+    /// as `-`, repeated hyphens collapsed, a `$esc` suffix dropped — so every
+    /// spelling the resolver accepts as UTF-8 (`utf_8`, `UTF8`, `utf8mb4`, …)
+    /// is accepted here too. A UTF-8 alias unknown to this list fails safe:
+    /// the literal parameter is used, nothing is misread.
     private static func isUTF8Label(_ charset: String) -> Bool {
-        let label = charset.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
-        return ["utf-8", "utf8", "utf8mb4", "us-ascii", "ascii", "iso646-us"].contains(label)
+        var label = charset
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+            .trimmingCharacters(in: CharacterSet(charactersIn: "\"'"))
+            .lowercased()
+            .replacingOccurrences(of: "_", with: "-")
+        while label.contains("--") {
+            label = label.replacingOccurrences(of: "--", with: "-")
+        }
+        if label.hasSuffix("$esc") {
+            label = String(label.dropLast(4))
+        }
+        return ["utf-8", "utf8", "utf8mb4"].contains(label)
     }
 
     /// Collect the RFC 2231 §3 continuation sections of an extended
--- a/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
@@ -3,6 +3,7 @@
 // numbered continuation sections.
 
 import Foundation
+import SwiftCross
 
 extension EMLParser {
 
--- a/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
@@ -55,8 +55,21 @@ extension EMLParser {
             // still rejects a byte sequence that is not text.
             return String(bytes: bytes, encoding: .utf8)
         }
-        guard let encoding = String.Encoding(ianaCharsetName: charset) else { return nil }
+        guard let encoding = String.Encoding(ianaCharsetName: charset),
+              encoding != .utf8 || isUTF8Label(charset) else { return nil }
         return String(bytes: bytes, encoding: encoding)
+    }
+
+    /// Whether a charset label names UTF-8 itself (or its US-ASCII subset).
+    ///
+    /// On platforms without CoreFoundation, `String.Encoding(ianaCharsetName:)`
+    /// resolves legacy charsets it has no converter for (GBK, Big5, EUC-KR,
+    /// KOI8-R, …) to `.utf8` as a best-effort placeholder. Here a wrong
+    /// decode is worse than none, because `nil` lets the sender's literal
+    /// spelling win, so a UTF-8 result is trusted only for a UTF-8 label.
+    private static func isUTF8Label(_ charset: String) -> Bool {
+        let label = charset.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        return ["utf-8", "utf8", "utf8mb4", "us-ascii", "ascii", "iso646-us"].contains(label)
     }
 
     /// Collect the RFC 2231 §3 continuation sections of an extended
--- a/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
+++ b/Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
@@ -42,50 +42,46 @@ extension EMLParser {
     /// `charset'language'` prefix names.
     ///
     /// RFC 2231 §4 makes the charset field optional — `filename*=''a.pdf`
-    /// is legal, and the value is the bytes as written. Any charset the
-    /// platform can name is honored, so a filename labelled `windows-1252`
-    /// or `iso-8859-15` decodes as such rather than being discarded; a
-    /// charset the platform does not know, or bytes that are not valid in
-    /// the charset named, yield `nil` and the literal spelling of the
-    /// parameter, if the sender wrote one, is used instead.
+    /// is legal — but leaving it blank "MUST NOT be done in order to
+    /// indicate a default character set", so only US-ASCII bytes have a
+    /// meaning there; anything else yields `nil`. Any charset the platform can name is honored, so a filename
+    /// labelled `windows-1252` or `iso-8859-15` decodes as such rather than
+    /// being discarded; a charset the platform does not know, or bytes that
+    /// are not valid in the charset named, yield `nil`. A `nil` result lets
+    /// the literal spelling of the parameter, if the sender wrote one, win.
     private static func decodeExtendedBytes(_ bytes: [UInt8], charset: String) -> String? {
         if charset.isEmpty {
-            // No charset was named, so the bytes carry no declared meaning
-            // beyond US-ASCII. UTF-8 reads every US-ASCII value unchanged and
-            // still rejects a byte sequence that is not text.
-            return String(bytes: bytes, encoding: .utf8)
+            guard bytes.allSatisfy({ $0 < 0x80 }) else { return nil }
+            return String(bytes: bytes, encoding: .ascii)
         }
         guard let encoding = String.Encoding(ianaCharsetName: charset),
-              encoding != .utf8 || isUTF8Label(charset) else { return nil }
+              encoding != .utf8 || isGenuineUTF8Label(charset) else { return nil }
         return String(bytes: bytes, encoding: encoding)
     }
 
-    /// Whether a charset label names UTF-8 itself.
+    /// Whether a `.utf8` resolution of `charset` names UTF-8 rather than
+    /// standing in for a charset the platform cannot decode.
     ///
-    /// On platforms without CoreFoundation, `String.Encoding(ianaCharsetName:)`
-    /// resolves legacy charsets it has no converter for (GBK, Big5, EUC-KR,
-    /// KOI8-R, …) to `.utf8` as a best-effort placeholder. Here a wrong
-    /// decode is worse than none, because `nil` lets the sender's literal
-    /// spelling win, so a `.utf8` result is trusted only for a label that
-    /// really names UTF-8. The label is normalized the way the resolver
-    /// normalizes it before lookup — trimmed, unquoted, lowercased, `_` read
-    /// as `-`, repeated hyphens collapsed, a `$esc` suffix dropped — so every
-    /// spelling the resolver accepts as UTF-8 (`utf_8`, `UTF8`, `utf8mb4`, …)
-    /// is accepted here too. A UTF-8 alias unknown to this list fails safe:
-    /// the literal parameter is used, nothing is misread.
-    private static func isUTF8Label(_ charset: String) -> Bool {
-        var label = charset
-            .trimmingCharacters(in: .whitespacesAndNewlines)
-            .trimmingCharacters(in: CharacterSet(charactersIn: "\"'"))
-            .lowercased()
-            .replacingOccurrences(of: "_", with: "-")
-        while label.contains("--") {
-            label = label.replacingOccurrences(of: "--", with: "-")
-        }
-        if label.hasSuffix("$esc") {
-            label = String(label.dropLast(4))
-        }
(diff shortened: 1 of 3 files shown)


Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
```diff
@@ -0,0 +1,162 @@
...
+    /// the charset named, yield `nil` and the literal spelling of the
+    /// parameter, if the sender wrote one, is used instead.
+    private static func decodeExtendedBytes(_ bytes: [UInt8], charset: String) -> String? {
+        if charset.isEmpty {
+            // No charset was named, so the bytes carry no declared meaning
+            // beyond US-ASCII. UTF-8 reads every US-ASCII value unchanged and
+            // still rejects a byte sequence that is not text.
+            return String(bytes: bytes, encoding: .utf8)
+        }
+        guard let encoding = String.Encoding(ianaCharsetName: charset) else { return nil }
```
odrobnik: **MEDIUM** — Do not treat unsupported legacy charsets as UTF-8. SwiftCross 1.2.0's non-Apple resolver deliberately maps unsupported euc-kr, gb2312, gbk, gb18030, big5, koi8-r, and macintosh labels to .utf8 rather than nil, so this code mistakes a placeholder for a real decoder. On Linux, `attachment; filename="fallback.txt"; filename*=gbk''%C2%A3.txt` is returned as `£.txt` because C2 A3 is valid UTF-8, although those bytes represent `拢` in GBK; the incorrect extended value then outranks and suppresses `fallback.txt`. Before this change, GBK returned nil and the literal fallback won. Use a resolver that reports actual converter support or reject SwiftCross's UTF-8 placeholder mappings.

odrobnik (changes requested):
Requesting changes for one cross-platform RFC 2231 decoding regression: on non-Apple platforms, the new charset path can treat unsupported legacy encodings as UTF-8, corrupting valid filenames and suppressing the safe literal fallback. No builds or tests were run, per the review constraints.

- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+RFC2231.swift:58`

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
tabmail-kmyi: Accurate, thanks. Fixed in 9e84bfb: `decodeExtendedBytes` now accepts a `.utf8` resolution only when the label itself names UTF-8 or US-ASCII, so SwiftCross's placeholder for GBK/Big5/EUC-KR/KOI8-R/… yields `nil` and the literal `filename=` wins. Platforms with a real converter (Darwin) still decode GBK as GBK. Added `extendedParameterLegacyCharsetIsNeverMisreadAsUTF8`, which pins on every platform that the UTF-8 misread is never the result.

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
```diff
@@ -0,0 +1,175 @@
...
+    /// parameter, if the sender wrote one, is used instead.
+    private static func decodeExtendedBytes(_ bytes: [UInt8], charset: String) -> String? {
+        if charset.isEmpty {
+            // No charset was named, so the bytes carry no declared meaning
+            // beyond US-ASCII. UTF-8 reads every US-ASCII value unchanged and
+            // still rejects a byte sequence that is not text.
+            return String(bytes: bytes, encoding: .utf8)
+        }
+        guard let encoding = String.Encoding(ianaCharsetName: charset),
+              encoding != .utf8 || isUTF8Label(charset) else { return nil }
```
odrobnik: **MEDIUM** — SwiftCross 1.2.0 canonicalizes charset labels before lookup, including replacing `_` with `-` and collapsing repeated hyphens, but `isUTF8Label` only trims and lowercases the raw label. Therefore `filename*=utf_8''r%C3%A9sum%C3%A9.pdf` resolves successfully to `.utf8` and is then rejected by this guard; without a literal parameter the attachment loses its filename (and may receive a generated one), and with a literal parameter the lower-fidelity fallback incorrectly wins. This violates the PR requirement to honor every charset the platform resolver names. Normalize the allow-check exactly as the resolver does, or reject only the known legacy labels that SwiftCross maps to the UTF-8 placeholder.

odrobnik (changes requested):
The earlier cross-platform legacy-charset corruption is addressed at the new head, and the continuation work remains linear. Requesting changes for one remaining RFC 2231 charset regression: the new UTF-8 safeguard applies narrower label normalization than SwiftCross itself, so resolver-supported UTF-8 aliases are discarded. No builds or tests were run, per the review constraints.

- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+RFC2231.swift:59`

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
tabmail-kmyi: Accurate, thanks. Fixed in 53f27c9: `isUTF8Label` now normalizes exactly as the resolver does (trim, unquote, lowercase, `_`→`-`, collapse repeated hyphens, drop `$esc`) before matching `utf-8`/`utf8`/`utf8mb4`, and the dead US-ASCII entries are gone (they resolve to `.ascii`, never `.utf8`). New test `extendedParameterUTF8AliasesAreHonored` covers utf-8, UTF-8, utf8, UTF8, utf8mb4, utf_8, utf--8 and utf-8$esc on every platform; it is red on the previous guard for the last three.

tabmail-kmyi:
Both findings are addressed and the head is now 53f27c9:

- 9e84bfb: a `.utf8` resolution is accepted only when the label itself names UTF-8 or US-ASCII, so SwiftCross's placeholder for GBK/Big5/EUC-KR/KOI8-R/macintosh no longer decodes those bytes as UTF-8.
- 53f27c9: `isUTF8Label` normalizes exactly as the resolver does (trim, unquote, lowercase, `_` → `-`, collapse repeated hyphens, drop the escape prefix) before matching.

Ready for another look whenever you have a moment.

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
```diff
@@ -0,0 +1,191 @@
...
+    /// or `iso-8859-15` decodes as such rather than being discarded; a
+    /// charset the platform does not know, or bytes that are not valid in
+    /// the charset named, yield `nil` and the literal spelling of the
+    /// parameter, if the sender wrote one, is used instead.
+    private static func decodeExtendedBytes(_ bytes: [UInt8], charset: String) -> String? {
+        if charset.isEmpty {
+            // No charset was named, so the bytes carry no declared meaning
+            // beyond US-ASCII. UTF-8 reads every US-ASCII value unchanged and
+            // still rejects a byte sequence that is not text.
+            return String(bytes: bytes, encoding: .utf8)
```
odrobnik: **MEDIUM** — Do not decode an omitted charset as UTF-8 for non-ASCII bytes. RFC 2231 §4 says leaving the charset blank must not indicate a default charset, but this branch accepts every valid UTF-8 sequence. With `attachment; filename="fallback.txt"; filename*=''%C2%A3.txt`, SwiftMail returns `£.txt` and suppresses `fallback.txt`, even though the undeclared bytes could instead be GBK `拢` or Latin-1 `Â£`. Restrict the blank-charset path to unambiguous US-ASCII bytes (or return nil) so ambiguous bytes reach the literal fallback.

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
```diff
@@ -0,0 +1,191 @@
...
+        }
+        guard let encoding = String.Encoding(ianaCharsetName: charset),
+              encoding != .utf8 || isUTF8Label(charset) else { return nil }
```
odrobnik: **MEDIUM** — Do not apply the non-Apple UTF-8-placeholder allowlist to authoritative CoreFoundation resolutions. On Darwin, `String.Encoding(ianaCharsetName: "unicode-1-1-utf-8")` resolves to `.utf8`, but `isUTF8Label` rejects that legitimate alias. Thus `attachment; filename="fallback.pdf"; filename*=unicode-1-1-utf-8''r%C3%A9sum%C3%A9.pdf` returns `fallback.pdf` (or loses the filename when no literal is present) instead of `résumé.pdf`, violating the PR's promise to honor any charset the platform resolver names.

odrobnik (changes requested):
Requesting changes for two remaining RFC 2231 charset regressions. Blank charset fields currently assign UTF-8 meaning to otherwise ambiguous non-ASCII bytes and can suppress the safe literal fallback. Separately, the UTF-8 placeholder safeguard is applied on Darwin too, where it rejects legitimate aliases that CoreFoundation resolves to UTF-8. The continuation collection change is linear and the previously reported unsupported-legacy-label and normalization cases are addressed. No builds or tests were run, per the review constraints.

- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+RFC2231.swift:56`
- **MEDIUM** `Sources/SwiftMail/MIME/EMLParser+RFC2231.swift:59`

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
tabmail-kmyi: Accurate, thanks. Fixed in 9f1a2b6: with the charset field blank, `decodeExtendedBytes` now decodes only US-ASCII bytes (RFC 2231 §4: leaving the field blank "MUST NOT be done in order to indicate a default character set"); any byte ≥ 0x80 yields `nil`, so `filename*=''%C2%A3.txt` falls through to `fallback.txt`, or to no name when there is no literal. New test `extendedParameterWithBlankCharsetDecodesOnlyASCII` pins both outcomes plus `''a%20b.pdf` → `a b.pdf` and `'en'invoice.pdf` → `invoice.pdf`; it is red on the previous head.

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
tabmail-kmyi: Accurate, thanks. Fixed in 9f1a2b6: the placeholder guard is now compiled only under the exact platform condition SwiftCross uses for its hand table (`canImport(CoreFoundation) && (os(macOS) || os(iOS) || …)`); on Apple platforms every `.utf8` CoreFoundation returns is trusted, so `unicode-1-1-utf-8` decodes. On the non-Apple side the guard no longer mirrors the resolver's normalization: given the resolver already returned `.utf8`, the label's letters and digits must start with `utf8`, which every accepted UTF-8 spelling does and no placeholder label (gbk, big5, euc-kr, koi8-r, macintosh, and their aliases) can. `extendedParameterUTF8AliasesAreHonored` now includes `unicode-1-1-utf-8` on Darwin and asserts the literal fallback for it elsewhere; `extendedParameterLegacyCharsetIsNeverMisreadAsUTF8` now loops over every placeholder label and alias in SwiftCross's table. The non-Apple branch was checked against SwiftCross 1.2's tables by inspection; I have no Linux toolchain locally, so your CI is its first run.

tabmail-kmyi:
Both findings are addressed and the head is now 9f1a2b6:

- A blank charset field decodes only US-ASCII bytes; anything else yields `nil` and the literal `filename=` wins (RFC 2231 §4: a blank field "MUST NOT be done in order to indicate a default character set").
- The UTF-8 placeholder guard is compiled only where SwiftCross uses its hand table, under SwiftCross's own platform condition, so CoreFoundation's `.utf8` resolutions such as `unicode-1-1-utf-8` are trusted on Darwin. On other platforms the guard is a simple discriminator (label letters+digits start with `utf8`) rather than a copy of the resolver's normalization, so it cannot drift from it again.

627 tests in 77 suites pass, SwiftLint strict is clean, and the new assertions are red on the previous head. Ready for another look whenever you have a moment.

odrobnik:
MissionControl review timed out for head `9f1a2b6c875b`. Auto-merge remains blocked; human review is required.

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
```diff
@@ -0,0 +1,187 @@
...
+    /// labelled `windows-1252` or `iso-8859-15` decodes as such rather than
+    /// being discarded; a charset the platform does not know, or bytes that
+    /// are not valid in the charset named, yield `nil`. A `nil` result lets
+    /// the literal spelling of the parameter, if the sender wrote one, win.
+    private static func decodeExtendedBytes(_ bytes: [UInt8], charset: String) -> String? {
+        if charset.isEmpty {
+            guard bytes.allSatisfy({ $0 < 0x80 }) else { return nil }
+            return String(bytes: bytes, encoding: .ascii)
+        }
+        guard let encoding = String.Encoding(ianaCharsetName: charset),
```
odrobnik, line 58: **HIGH** — The new file imports only Foundation, but `String.Encoding.init?(ianaCharsetName:)` is declared by SwiftCross. Swift imports are file-scoped, there is no SwiftCross re-export in SwiftMail, and the existing files using this initializer explicitly import SwiftCross. Compiling the SwiftMail target therefore fails at this call on every platform before the RFC 2231 code or tests can run. Add `import SwiftCross` to this file.

odrobnik (changes requested):
Requesting changes for one build-blocking issue introduced by the RFC 2231 file split. The new file calls SwiftCross's IANA charset initializer without importing SwiftCross, so the SwiftMail target cannot compile. The continuation implementation is linear and the previously reported charset regressions are addressed. No builds or tests were run, per the review constraints.

- **HIGH** `Sources/SwiftMail/MIME/EMLParser+RFC2231.swift:57`

Sources/SwiftMail/MIME/EMLParser+RFC2231.swift
tabmail-kmyi: Thanks. The import was indeed missing and is added in 5e234fd. One correction to the consequence: the target did compile on every platform (the PR checks for 9f1a2b6 are green on macOS, iOS, Linux and Android), because Swift resolves extension members through any file in the module that imports the declaring module. That is exactly the leaky behaviour `MemberImportVisibility` turns into an error, so the explicit import is the right fix regardless.

tabmail-kmyi:
Addressed; head is now 5e234fd. `EMLParser+RFC2231.swift` imports SwiftCross explicitly. No behaviour change: 627 tests pass locally and SwiftLint strict is clean.

odrobnik (approved):
Approved. The previously reported charset and import issues are addressed at 5e234fd. I traced filename extraction and Content-Type cleanup through their parser call sites and checked SwiftCross 1.2.0's on-disk IANA resolver: the platform split and UTF-8 placeholder safeguard now match the dependency, blank charsets fail safely for non-ASCII bytes, and continuation collection remains linear with contiguous decimal semantics. No builds or tests were run, per the review constraints.

👍 No findings.

PR #234, opened 2026-09-10
Require the fetched message count before indexing in header fallback tests

tabmail-kmyi:
## Summary

`FetchMessageInfoHeaderFallbackTests` asserts the fetched count with a non-fatal `#expect(infos.count == 1)` and then immediately subscripts `infos[0]`. If the fetch ever returns nothing, the subscript traps and takes the whole test process down, hiding every other result instead of reporting one failed test.

This switches the six count checks to `try #require(infos.count == 1)`, matching the pattern already used in `EMLByteLevelTests`. No behavior change.

## Verification

- `swift test --filter FetchMessageInfoHeaderFallbackTests`: 6 tests in 1 suite passed.
- `swiftlint lint --strict`: clean.
- Fork CI (macOS, Linux, iOS, Android, SwiftLint) green on this exact commit.

Noticed while working on #233; kept separate since it is unrelated test hygiene.

### DIFF
--- a/Tests/SwiftIMAPTests/FetchMessageInfoHeaderFallbackTests.swift
+++ b/Tests/SwiftIMAPTests/FetchMessageInfoHeaderFallbackTests.swift
@@ -23,7 +23,7 @@ struct FetchMessageInfoHeaderFallbackTests {
             "A001 OK FETCH completed\r\n"
         ])
 
-        #expect(infos.count == 1)
+        try #require(infos.count == 1)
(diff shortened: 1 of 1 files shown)


odrobnik (approved):
No findings. The six changes correctly replace non-fatal count expectations with fatal `try #require` preconditions before `infos[0]` access. All affected tests already support throwing, the pattern is established in the same test target, and no production behavior is changed.

👍 No findings.

